PR #26, opened 2019-05-30
Simplify steps by using powershell on both platforms to install built package.

nelsonjchen:
Pro:

1 step instead of 2 for all platforms. Very consistent. The steps file is getting a bit bloated.

Con:

People don't usually use Powershell as their shell on macOS and Linux. The fact it is preinstalled on all Azure Pipeline runners on all the platforms is quite unusual. 

I don't know what decision to make. It's something I noticed.

### DIFF
--- a/ci/steps.yml
+++ b/ci/steps.yml
@@ -18,16 +18,10 @@ steps:
     displayName: build package
   #
   # Install aqtinstall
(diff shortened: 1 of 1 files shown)


miurahr:
Looks good. Thanks.

PR #27, opened 2019-05-30
Test mirror functionality

nelsonjchen:
Built on top of #25. 

Edited the generator script to specify or not specify a base URL mirror.

I don't know if I'm using this right or if I've set the wrong mirror URL somehow. `-b` doesn't work. I haven't debugged this yet though.

### DIFF
--- a/generate_azure_pipelines_yml.py
+++ b/generate_azure_pipelines_yml.py
@@ -86,13 +86,15 @@ windows_build_jobs.extend(
     ]
 )
 
-# All Androids for all platforms
+# Androids for Linux platforms
+# aqt is for CI/CD systems!
+# Users might develop on Win/Mac, but are most likely to use Linux for CI/CD with
+# the Android ecosystem.
 
-for android_arch in [ 'android_x86', 'android_armv7', ]:
-    for platform_build_jobs in all_platform_build_jobs:
-        platform_build_jobs.build_jobs.append(
-            BuildJob('5.13.0', platform_build_jobs.platform, 'android', android_arch, android_arch)
-        )
+for android_arch in ['android_x86', 'android_armv7']:
+    linux_build_jobs.append(
+        BuildJob('5.13.0', 'linux', 'android', android_arch, android_arch)
+    )
 
 matrices = {}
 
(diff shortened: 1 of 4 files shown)


miurahr:
Argument for '-b' option should  be 'http://mirrors.ocf.berkeley.edu/qt/'    instead of  'http://mirrors.ocf.berkeley.edu/qt/online/qtsdkrepository/'

generate_azure_pipelines_yml.py
```diff
@@ -42,6 +43,11 @@ def __init__(self, platform, build_jobs):
     PlatformBuildJobs('windows', windows_build_jobs),
 ]
 
+mirrors = [
+    None,
+    'http://mirrors.ocf.berkeley.edu/qt/online/qtsdkrepository/'
```
miurahr: it should be   'http://mirrors.ocf.berkeley.edu/qt/'

miurahr:
I've just pushed fix. @nelsonjchen could you pull/fetch your repository from github before further modifications?

miurahr:
Now tests are working, but it seems  too much testing the project.
Does @nelsonjchen think? 

Do we run test case such as
1. Windows  5.11.3 desktop  mirror = some
2. Windows  5.11.3  android  mirror = None
etc..
to reduce number of test case a half.

nelsonjchen:
I agree, it's too much. There are some other jobs that could be dropped too.

nelsonjchen:
People building Android apps usually just build on Linux; no Mac or Windows needed. 

If we're testing mirror functionality, I think it can just be a separate job by itself, not parameterized across the current set of jobs.

We can also move lint out to its own job on Linux. Linting's quick, but it would also help cut down that large `steps.yml` file to just building QT and testing.

miurahr:
>We can also move lint out to its own job on Linux. Linting's quick, but it would also help cut down that large steps.yml file to just building QT and testing.

I agree on it. I've already cut it on master branch. lint job is done on Travis-CI where I put a release automation script.

miurahr:
Merged. Thanks a lot!!

Issue #7, opened 2019-03-04
Replace 7zip with LZMA standard library to drop external dependency


miurahr:

miurahr:
LZMA supports LZMA lossless compression algorithm and xz file but  not support 7z archive format.

miurahr:
https://github.com/fancycode/pylzma/blob/master/py7zlib.py

miurahr:
Waiting external activity

miurahr:
https://github.com/fancycode/pylzma has just updated to support recent compress formats.

miurahr:
Now working on topic branch   https://github.com/miurahr/aqtinstall/tree/topic-py7zr

PR #31, opened 2019-06-01
Use py7zr for extracting package 7z archive

miurahr:
Implement #7


miurahr:
It may be better to add a command line  option  to select method to extract among py7zr, libarchive, 7-zip and p7zip.

PR #32, opened 2019-06-01
Use Azure Pipelines itself to generate and run test matrix from Python script

### LINKED ISSUE #30: Look into azure-pipelines self-generated matrix from a script
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-a-multi-job-output-variable

The PRs I have open now are a bit of a mess because I haven't had time to sort out the merge conflicts in the generated artifacts. 

It's not good to commit generated artifacts. 

But if I implement the above linked approach, we should be able to get rid of the generated yaml workflow that must be done pre-commit.

nelsonjchen:
Instead of having to run a script on the developer's system and then committing the result, Azure Pipelines itself can run a modification of that script and reuse the variable it sets in later steps to determine the test matrices.

Reference: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-configuration

With this, `/azure-pipelines.yml` no longer needs to be treated like a committed generated artifact!

Closes #30.


miurahr:
Thanks!  A failure is not related to this PR.

PR #56, opened 2019-11-16
Support explicit spec for extra modules

Skycoder42:
I took your work from the `topic-extra-component` branch and added on to it in hopes to get a working configuration for extra modules. I also squashed the two commits into one for a cleaner history. If you prefer to keep the separated, I can change that.


miurahr:
Thank you!

PR #60, opened 2019-11-17
Add extended android support

Skycoder42:
This adds new combinations to support the x64 versions of android (x86_64 and arm64_v8a). For !t 5.14, all android plattforms have been combined into one multiarch platform, simply named "android". This one has been added as well.


aqt/cli.py
```diff
@@ -83,6 +83,8 @@ def _set_arch(self, args, oarch, os_name, target):
                 arch = "clang_64"
             elif os_name == "mac" and target == "ios":
                 arch = "ios"
+            elif target == "android":  # TODO only works for Qt >= 5.14. Somehow parse the version?
```
Skycoder42: As you can see, we theoretically have to check if `qt_version` is >= 5.14 to prevent user errors. However, mentioning this in the README might be enough. What do you prefer:

- Using a version comparison like: https://packaging.pypa.io/en/latest/version/#packaging.version.parse
- By adding a hint to the readme

Skycoder42:
@miurahr Regarding my comment above: Which variant do you prefer? If you don't have a preference, I'd go with a version compare

aqt/cli.py
miurahr: I prefer both. 

When using version comparison and set default values.
When 5.14.0 and later, set 'android' and when 5.13.x and prior set 'android_armv7'.
If a default value is natural, we can set default. If not, it should be explicitly specified as same as MSVC arch parameter.

miurahr:
Could you also update following section in README ?

- For android and windows you also need to specify an arch: win64_msvc2017_64, win64_msvc2015_64, win32_msvc2015, win32_mingw53, win64_mingw73, android_x86, android_armv7

miurahr:
Fails in travis-CI  is caused by master's issue, not the PR's.

README.rst
```diff
@@ -62,8 +62,12 @@ General usage looks like this:
 * The Qt version is formatted like this: `5.11.3`
 * Host is one of: `linux`, `mac`, `windows`
 * Target is one of: `desktop`, `android`, `ios` (iOS only works with mac host)
-* For android and windows you also need to specify an arch: `win64_msvc2017_64`,
-  `win64_msvc2015_64`, `win32_msvc2015`, `win32_mingw53`, `win64_mingw73`, `android_x86`, `android_armv7`
+* For some platforms you also need to specify an arch:
+  * For windows, choose one off: `win64_msvc2017_64`, `win32_msvc2017`, `win64_msvc2015_64`, `win32_msvc2015`, 
+  `win64_mingw73`, `win32_mingw73`, `win64_mingw53`, `win32_mingw53`, `win64_msvc2017_winrt_x64`, 
+  `win64_msvc2017_winrt_x86`, `win64_msvc2017_winrt_armv7`
+  * For android and Qt 5.13 or below, choose one off: `android_x86_64`, `android_arm64_v8a`, `android_x86`,
```
miurahr:
```
- choose one off
+ choose one of
```

README.rst
```diff
@@ -62,8 +62,12 @@ General usage looks like this:
 * The Qt version is formatted like this: `5.11.3`
 * Host is one of: `linux`, `mac`, `windows`
 * Target is one of: `desktop`, `android`, `ios` (iOS only works with mac host)
-* For android and windows you also need to specify an arch: `win64_msvc2017_64`,
-  `win64_msvc2015_64`, `win32_msvc2015`, `win32_mingw53`, `win64_mingw73`, `android_x86`, `android_armv7`
+* For some platforms you also need to specify an arch:
+  * For windows, choose one off: `win64_msvc2017_64`, `win32_msvc2017`, `win64_msvc2015_64`, `win32_msvc2015`,
```
miurahr: choose one off > one of

aqt/combinations.json
```diff
@@ -1,6 +1,9 @@
 [{"qt": [
   {"os_name": "linux",   "target": "desktop", "arch": "gcc_64"},
   {"os_name": "linux",   "target": "desktop", "arch": "wasm_32"},
+  {"os_name": "linux",   "target": "android", "arch": "android"},
+  {"os_name": "linux",   "target": "android", "arch": "android_x86_64"},
+  {"os_name": "linux",   "target": "android", "arch": "android_arm64_v8a"},
```
miurahr: Is it any way to specify version range in the database? If we can do it, we can check it.
For example, "version_maximum": "5.13", "version_minimum":"5.14"

It can be also useful for WASM module.

aqt/cli.py
```diff
@@ -27,6 +27,7 @@
 import os
 import platform
 import sys
+from packaging.version import Version, parse
```
miurahr: packaging seems a known third party library.
It is better to add it to tox.ini in section isort;
```
[isort]
known_first_party = aqt
known_third_party = docutils,flake8,pyannotate_runtime,pytest,pytz,setuptools,sphinx,yaml
```
and change import  order,
```
 import os
 import platform
 import sys
+
+import yaml
 from packaging.version import Version, parse
-
-import yaml
```

aqt/combinations.json
Skycoder42: Yes, this is possible. However, I have to admit I don't think this database is a good idea at all. Let the user do whatever combinations they like and print an error if a combination does not exist. Or at least add a parameter to make the database check a warning instead of an error, so one can test combinations in alphas of Qt etc. that are not covered by the database yet.

miurahr: OK, i'd like to merge it as is.

PR #85, opened 2020-02-07
replace scripts with console_scripts entry_point

graingert:


miurahr:
Please fix a failure of CI test.
You may need to remove bin and bin/aqt file.

PR #96, opened 2020-02-22
Allow path search for 7z

altendky:


altendky:
The existing code did not search `$PATH` and would reject just `7z` even when it would run fine later on.  My first attempt with `shutil.which()` isn't really great.  Just try, let it fail, catch the exception.  A single subprocess call isn't a huge overhead in this context.

altendky:
This is relevant to using an external 7z command across platforms where paths would vary but it may always be findable in `$PATH`.

miurahr:
Looks good.

miurahr:
Also picked for 0.7.*

PR #97, opened 2020-02-26
No more os.chdir()

altendky:
WIP for:
- [x] requiring fixed release version of py7zr
- [x] Testing together in CI
- [x] ~~May as well update for the pywin32 dep of py7zr as well when there's a (beta) release~~
  - This should happen before releasing aqtinstall but isn't related to this PR...
  - Well, other than this PR failing so, sure, depends on preference, but it needs to happen regardless


altendky:
Pairs with miurahr/py7zr#82

PR #112, opened 2020-03-07
Support external settings

miurahr:


miurahr:
Restart effort. Drop here.

PR #140, opened 2020-05-31
WIP: Introduce  debug log options

miurahr:
Introduce -q --quiet and --verbose, -v, -vv options


miurahr:
When verbose option specified, it set logger.setLevel() which is change a verbosity of root of message flow.
logging.ini set structure for file and console levels. command line option should change console log level only.

Issue #149, opened 2020-07-31
Qt5.7 installation failed: Download error: decompression get wrong: no output data


pbek:
I was sent over from https://github.com/jurplel/install-qt-action/issues/50 with an install problem:

I'm currently having issues with the Qt 5.7  installation on Windows (some still are using Windows XP) with `win32_mingw53`.

Config: https://github.com/pbek/QOwnNotes/blob/48f5b73e1b0a200b88c93d82ef6192f7d80fbe03/.github/workflows/build.yml#L61-L69
Build log: https://github.com/pbek/QOwnNotes/runs/880829548?check_suite_focus=true

I always get the error `Download error: decompression get wrong: no output data.` for all packages:

```
2020-07-17 06:56:23,560 - aqt - INFO - Downloading qtsvg...
2020-07-17 06:56:23,560 - aqt - DEBUG - Download URL: http://mirrors.ocf.berkeley.edu/qt//online/qtsdkrepository/windows_x86/desktop/qt5_57/qt.57.win32_mingw53/5.7.1-0qtsvg-Windows-Windows_7-Mingw53-Windows-Windows_7-X86.7z
2020-07-17 06:56:24,064 - aqt - ERROR - Download error: decompression get wrong: no output data.
```

- I tried with and without mirror
- Qt 5.14.2 works great, only Qt 5.7 doesn't seem to work
- When I download the file on my computer it decompresses fine

Has anyone seen something like that?

miurahr:
This caused by dependency project py7zr's fixed bug
https://github.com/miurahr/py7zr/issues/178
https://github.com/miurahr/py7zr/releases/tag/v0.9.0b1


Could you specify py7zr version `py7zrversion` as v0.9.0b3 in install-qt-action?

pbek:
> Could you specify py7zr version py7zrversion as v0.9.0b3 in install-qt-action?

Thank you for your quick answer, I'll ask back!

pbek:
py7zr 0.9 was installed: https://github.com/pbek/QOwnNotes/runs/930611080?check_suite_focus=true#step:4:29

But I still get the same error: https://github.com/pbek/QOwnNotes/runs/930611080?check_suite_focus=true#step:4:80

miurahr:
py7zr v0.9 was installed,  but uninstalled and intalled again v0.7 later.
```
Installing collected packages: six, pyparsing, packaging, certifi, urllib3, idna, chardet, requests, py7zr, aqtinstall
[...]

PR #150, opened 2020-07-31
Fix installation of Qt5.7 which use LZMA1+BCJ

miurahr:
Update dependency to v0.9.0 and later.
Fix for #149

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ install_requires =
     importlib-metadata;python_version<"3.8"
     requests
     wheel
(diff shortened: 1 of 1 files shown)


miurahr:
Extraction performance is significantly decreased. Need investigation.

miurahr:
Fix for performance pushed.

Ubuntu py38 Qt5.15 resulted from 8sec with py7zr 0.7.4 -> 10sec with py7zr 0.9.0, 
which would be acceptable range.

PR #159, opened 2020-09-09
Add check on packages name to download

Gamso:
When we start an installation, all packages are downloaded whatever the specified platform.


Gamso:
For more details, 

i used this command line "aqt install --outputdir c:\Qt windows desktop win32_msvc2019 -m all".
Like explained in the documentation, i use "-m all" to get all extra modules. 

Into class "QtArchive", if self.all_extra = True, the method "_parse_update_xml()" accept all packages ...

Gamso:
I hope that the test is enough strong. I was a bit confused about test syntax x) (first time i did a python test).

miurahr:
@Gamso Thank you for you help, but... could you add test data to repository?
You may forget to run `git add ...` command.

tests/test_archives.py
```diff
@@ -20,3 +20,35 @@ def _mock(self, url):
...
+
+    def _mock(self, url):
+        with open(os.path.join(os.path.dirname(__file__), 'data', datafile), 'r') as f:
+            self.update_xml_text = f.read()
+
+    monkeypatch.setattr(QtArchives, "_download_update_xml", _mock)
+
+    # Get packages without extra modules
+    qt_archives = QtArchives(os_name, 'desktop', version, target)
+    assert qt_archives.archives is not None
```
miurahr, line 40: This assertion is as same as `test_parse_update_xml()`. 
It is better to merge two tests: `test_parse_update_xml()` and `test_parse_update_xml_with_all_extra_modules` into single `test_parse_update_xml()`.

tests/test_archives.py
```diff
@@ -20,3 +20,35 @@ def _mock(self, url):
...
+
+    # Extract all urls
+    url_list = [item.url for item in qt_archives.archives]
+    url_all_modules_list = [item.url for item in qt_archives_all_modules.archives]
+
+    # Check the difference list contains only extra modules urls for target specified
+    list_diff = [item for item in url_all_modules_list if item not in url_list]
+
+    # Assert if list_diff contains urls without target specified
+    assert len([item for item in list_diff if target not in item]) == 0
```
miurahr, line 54: It is difficult to debug if difference happened.
it would be better assert with null list `[]` then test error message show its contents.

for example,
```
unwanted_targets = [item for item in list_diff if target not in item]
assert unwanted_targets == []
```

miurahr (changes requested):
I now understand what you want to fix.
It is better to reduce a duplication not to test same thing twice, and to improve assertion.

miurahr:
Thanks @Gamso !

PR #177, opened 2020-11-20
add qt 5.15.2

Milerius:
https://www.qt.io/blog/qt-5.15.2-released


Milerius:
@miurahr the timeout doesn't seems to be my fault

PR #182, opened 2020-11-24
When we used "-m all" to download doc or examples, Qt sources are also downloaded

Gamso:
Add filter on doc and example packages when option "all_extra" is enabled.
Add test.

### DIFF
--- a/tests/test_doc_archives.py
+++ b/tests/test_doc_archives.py
@@ -1,0 +1,35 @@
+import os
+
+import pytest
+
+from aqt.archives import QtArchives
+
+
+@pytest.mark.parametrize("os_name,version,target,datafile", [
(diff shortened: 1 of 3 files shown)


tests/test_doc_archives.py
```diff
@@ -0,0 +1,35 @@
...
+
+import pytest
+
+from aqt.archives import QtArchives
+
+
+@pytest.mark.parametrize("os_name,version,target,datafile", [
+    ('windows', '5.15.2', 'doc', 'windows-5152-src-doc-example-update.xml')
+])
+def test_parse_update_xml(monkeypatch, os_name, version, target, datafile):
```
miurahr, line 11: Please change test function name other than the name of test in test_archive.py.

miurahr: Uhm, it is not hazardous from TravisCI result.

Issue #194, opened 2021-01-19
ConnectionError "Connection reset by peer"


samfrances:
I'm getting the following error when running `aqt install -O $QT_INSTALL_DIR 5.12.4 linux desktop`. I'm getting this on multiple python versions (3.7 and 3.9), multiple machines, and from multiple different network locations.

Any help much appreciated.

```
$ aqt install -O $PWD 5.12.4 linux desktop
2021-01-19 11:44:03,071 - aqt - INFO - aqtinstall(aqt) v0.10.1 on Python 3.9.1 [CPython GCC 5.4.0 20160609]
2021-01-19 11:45:13,280 - aqt - ERROR - Download error: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
Traceback (most recent call last):
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 382, in _make_request
    self._validate_conn(conn)
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1010, in _validate_conn
    conn.connect()
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/connection.py", line 411, in connect
    self.sock = ssl_wrap_socket(
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 428, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
  File "/home/sam/temp/venv/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 472, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/local/lib/python3.9/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/local/lib/python3.9/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/local/lib/python3.9/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
ConnectionResetError: [Errno 104] Connection reset by peer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/sam/temp/venv/lib/python3.9/site-packages/requests/adapters.py", line 439, in send
[...]

PR #196, opened 2021-01-20
Select fallback mirror site when main site connection failed

miurahr:
Add feature to select fallback mirror site randomly from known sites.

ref #194


sionhughes:
I can't see this functionality in 0.11.1 release? Am I missing something?

miurahr:
https://github.com/miurahr/aqtinstall/blob/359375eb6a7fd94c991981bb525f69a5778a7714/aqt/installer.py#L60-L64

https://github.com/miurahr/aqtinstall/blob/359375eb6a7fd94c991981bb525f69a5778a7714/aqt/installer.py#L195-L203

PR #200, opened 2021-01-26
Update list command to work for Qt6

miurahr:


miurahr:
Failed test is not caused by here, so we can merge.

PR #225, opened 2021-03-15
Add --keep option to keep downloaded archives

miurahr:
Introduce '-k' or '--keep' command line option


jayvdb (approved):
I would like this, as it would help speed up test runs of a package I am adding tests for (qt-applications).

Issue #233, opened 2021-04-13
Installation of Qt IFW tools doesn't give working IFW tools


sionhughes:
Within Gitlab CI pipeline on multiple projects, I've witnessed installation of Qt IFW tools resulting in an invalid set of tools available on the build platform. The tools are used to build an installer via `cpack` of a Windows based Qt project. 

Error output via cpack as it can't find all parts of IFW toolset:
```
CPack Error: Cannot find QtIFW compiler "binarycreator": likely it is not installed, or not in your PATH
CPack Error: Cannot initialize the generator IFW
```

Installation of Qt IFW tools is done using the following command:
```
aqt tool --outputdir C:\Qt windows tools_ifw 4.0 qt.tools.ifw.40
```

On 13/04/21 @ 04:03 UTC, the installation took `9.16696960 second` (working state, installer built OK) 
On 13/04/21 @20:32 UTC, the installation took `1.10179330 second` (non-working state, running `cpack` results in error)

What could have changed for the installation? Change of sources from the Qt server?

sionhughes:
Just looking a little further in, looks like Qt IFW tools 4.1.0 was released today:

![image](https://user-images.githubusercontent.com/7159235/114620324-5edbd680-9ca3-11eb-8cb7-4ef8fa4ac073.png)

Could that have broken things?

sionhughes:
Right.... some more digging. 

Is this because the `qt.tools.ifw.40` dir is no longer present, and replaced by `qt.tools.ifw.41` at https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_ifw/ ?

Installing version 4.1 using `aqt tool --outputdir C:\Qt windows tools_ifw 4.1 qt.tools.ifw.41` locally on my laptop results in tools being installed. Made changes to CI pipelines now kicking them off to prove this change. Will report back in the morning as I'm not waiting for them to run now.

pzhlkj6612:
> Is this because the qt.tools.ifw.40 dir is no longer present, and replaced by qt.tools.ifw.41 at https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_ifw/ ?

Yes, it is. You can open [Updates.xml (linux_x64)](https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_ifw/Updates.xml) to see what package(s) can be installed.

miurahr:
There are many space to improve for the tools installation.
[...]

PR #235, opened 2021-05-02
Add capability to list tools

ddalcino:
This is an intermediate step towards addressing the issues in #233. This allows you to type `aqt list tools linux desktop`, and get a list of all the tools on the server. There are more options; type `aqt list -h` to see them all.

~~Something like this functionality should probably be folded into `aqt list`, but I'm not quite certain what the eventual interface should look like. Right now, this PR is only intended to demonstrate how aqt could go about listing all the tools available at downloads.qt.io.~~

~~Potentially, something like this could be used to tell the user what versions of Qt are available, or automatically grab the newest version of Qt instead of a predefined version.~~

There is a pretty big caveat here: This code will need to be altered if the Qt devs decide to redesign the downloads.qt.io page. To my knowledge, they have not done so in the 5 years I have been using Qt, but it's bound to happen eventually.


miurahr:
@ddalcino could you merge master here for fixing doc tests?
And could you add a test case for the enhancement?

miurahr:
@ddalcino could you extend `list` command for the feature?

When user specify `category host target` , aqt run the proposed feature and show versions and packages.
When user specify `category host target version` then list packages by reading Updates.xml.

For example, 

```
$ python -m aqt list qt5 linux desktop
5.9.0 5.9.1 ...
...
5.12.0 5.12.1 ..
..
5.15.0 5.15.1 5.15.2

$ python -m aqt list qt5 linux desktop 5.15.2
"qtcharts", "qtlottie", "qtnetworkauth", "qtpurchasing", "qtdatavis3d", "qtquick3d", "qtquicktimeline", "qtscript", "qtvirtualkeyboard", "qtwebengine", "qtwebglplugin"

```

ddalcino:
> @ddalcino could you extend `list` command for the feature?
> 
> When user specify `category host target` , aqt run the proposed feature and show versions and packages.
> When user specify `category host target version` then list packages by reading Updates.xml.

Yes, I think I can do this. 

I would like to suggest that we add options to show the `wasm`, `src_doc_examples`, `preview`, and `wasm_preview` directories as well. For example, something like:

    $ python -m aqt list qt5 linux desktop wasm
    5.13.0 5.13.1 ...

There's a `qt5_5130_wasm` directory Linux, but not on Mac or Windows. I imagine that there are other inconsistencies like this, and it might be useful to be able to discover them with a CLI tool like aqt.

On the other hand, I'm not too crazy about `aqt list <category> <os> <target> [wasm|src_doc_examples|preview|wasm_preview|<version>]`; as a CLI user, I might get confused by that syntax. Just a thought.

ddalcino:
> @ddalcino could you merge master here for fixing doc tests?

Ok, done!

> And could you add a test case for the enhancement?

I'm not sure about how to do test cases for something like this. The expected results are a moving target: Qt adds a new version every 6 months, and eventually they will remove the non-LTS versions. I will give it a try though.

miurahr:
> And could you add a test case for the enhancement?
> 
> I'm not sure about how to do test cases for something like this. The expected results are a moving target: Qt adds a new version every 6 months, and eventually they will remove the non-LTS versions. I will give it a try though.

You can add a scraping snapshot data as a test data, and make a stub which returns the  data instead of access a download site.

You can also add a test case to actually access the site like other cases in azure pipeline test. We can check whether aqt can get a recent versions such as 6.0.x and 5.15.x.

aqt/installer.py
```diff
@@ -646,12 +706,68 @@ def _create_parser(self):
...
+            "IE: `aqt list qt5 windows desktop --filter-minor 12` prints all versions beginning with 5.12",
+        )
+        output_modifier_exclusive_group = list_parser.add_mutually_exclusive_group()
+        output_modifier_exclusive_group.add_argument(
+            "--list-packages",
+            type=str,
+            metavar="VERSION",
+            help='Qt version in the format of "5.X.Y". '
+            "When set, this lists all the packages available for Qt 5.X.Y.",
+        )
```
ddalcino, line 759: > When user specify `category host target version` then list packages by reading Updates.xml.

@miurahr, this is the optional argument I am using to list packages from Updates.xml. Personally, I prefer to require a flag in front of optional arguments like this because I think it makes the interface easier to understand. If you disagree with this choice, please let me know and I will change it to what you originally asked for.

aqt/installer.py
```diff
@@ -646,12 +706,68 @@ def _create_parser(self):
...
+            "--latest-version",
+            action="store_true",
+            help="print only the newest version available",
+        )
+        output_modifier_exclusive_group.add_argument(
+            "--latest-packages",
+            action="store_true",
+            help="list all the packages available for the latest version of Qt, "
+            "or a minor version if the `--filter-minor` flag is set.",
+        )
```
ddalcino, line 770: These two options, along with the `--filter-minor` option, are intended to help people who might be using the output of `aqt list` in a shell script, and don't want to have to filter the output manually. I can take them out if necessary.

aqt/installer.py
```diff
@@ -472,7 +481,7 @@ def run_tool(self, args):
             )
         )
 
-    def run_list(self, args):
+    def run_list_old(self, args):
```
ddalcino, line 484: @miurahr, when you asked me to extend `aqt list`, I wasn't sure how to preserve the original implementation. You asked for:

> When user specify `category host target`, aqt run the proposed feature and show versions and packages.
> When user specify `category host target version` then list packages by reading Updates.xml.

You provided an expected output for the `category host target version` command, which is a list of packages. The existing implementation of `aqt list` prints a table that does not match the expected output that you asked for. I implemented `helper.get_packages_for_version`, which prints the list of packages that you asked for, and removed the existing implementation.

Please let me know if I misunderstood what you meant.

aqt/installer.py
miurahr: It looks better than my proposal. Please feel free to change UI because existent  list sub command is quite incomplete.

miurahr:
When you are ready for merge, could you run 'black' utility to format a code?
You can run a project lint by `tox -e check`
There seems a problem in a current github actions configuration, a linting seems not to run here.

aqt/installer.py
ddalcino: Ok, I can remove the existing implementation. I will also remove the aqt.archives.PackagesList class. It is only used in Cli.run_list, and when I remove that function, it becomes dead code.

ddalcino: I only have implementations for options 1 and 4. I will avoid removing aqt.archives.PackagesList; there must be a way to reuse it for options 2 and 3.

I think we need a well-defined interface for accessing options 1-4. I don't like the idea of reusing positional arguments to mean different things; this makes the help text difficult to read and the interface unclear. Consider this possible interface:
```
$ list <category> <host OS> <target>   # 1. Print a list of all tools/versions of Qt available for category, os, target
$ list <Qt version> <host OS>          # 2. Print a list of all targets for a version of Qt
$ list <Qt version> <host OS> <target> # 3. Print a list of all architectures available for Qt version, os, target
$ list <Qt version> <host OS> <target> --modules # 4. Print all modules for (Qt, OS, target) tuple
```
This could work, but the first positional argument is confusing. In the help text, you would have to call it something like `QT_VERSION_OR_CATEGORY`, and the help text would have a lot of explaining to do.

This is an interface that I would prefer, as a user. It is more verbose, but it's also more consistent.
```
$ list <category> <host OS> <target>   # 1. Print a list of all tools/versions of Qt available for category, os, target
$ list <category> <host OS> --targets <Qt version>       # 2. Print a list of all targets for a version of Qt
$ list <category> <host OS> <target> --arch <Qt version> # 3. Print a list of all architectures available for Qt version, os, target
$ list <category> <host OS> <target> --modules <Qt version> # 4. Print all modules for (Qt, OS, target) tuple
```

Another possibility is to turn option 1 into separate subcommands. This option is less verbose and easier to understand:

```
$ list-tools <host OS> <target> # 1.1. Print a list of all tools available for os, target
$ list-ver <host OS> <target>   # 1.2. Print a list of all versions of Qt available for category, os, target. 
                                # There could be flags for filtering by major and minor versions.
$ list <Qt version> <host OS>          # 2. Print a list of all targets for a version of Qt
$ list <Qt version> <host OS> <target> # 3. Print a list of all architectures available for Qt version, os, target
$ list <Qt version> <host OS> <target> --modules # 4. Print all modules for (Qt, OS, target) tuple
```

These are the best suggestions I can come up with right now. There are many other ways to solve this problem; maybe you can come up with something better?

miurahr: Thank you for detailed examination and design considerations! 🥇 

> This is an interface that I would prefer, as a user. It is more verbose, but it's also more consistent.

I prefer a middle No. 2️⃣ , too. Could you proceed your design? 🚀 

```
$ list <category> <host OS> <target>   # 1. Print all tools/versions of Qt available for category, os, target
$ list <category> <host OS> --targets <Qt version>       # 2. Print all targets for a version of Qt
$ list <category> <host OS> <target> --arch <Qt version> # 3. Print all architectures for Qt version, os, target
$ list <category> <host OS> <target> --modules <Qt version> # 4. Print all modules for (Qt, OS, target) tuple
```

ddalcino: Ok, number 2 it is!

miurahr:
Test failed because some mirror site is problematic 'mirrors.sjtug.sjtu.edu.cn'

miurahr:
It is better to update [ci/generate_azure_pipelines_matrices.py](https://github.com/miurahr/aqtinstall/blob/v1.2.0b2/ci/generate_azure_pipelines_matrices.py)to add a matrix of test, then add minimal steps to YAML file.

Current [ci/steps.yml](https://github.com/miurahr/aqtinstall/blob/v1.2.0b2/ci/steps.yml) has already been complex enough, and is necessary to refactor to keep maintainability....

ddalcino:
Ok, I'll try that.

I've never used Azure before, but I'm sure I can figure it out. Thanks for your help!

miurahr:
I worked for refactoring azure-pipeline script at #250.
you can add list command test case 
https://github.com/miurahr/aqtinstall/blob/topic-refactor-azure/ci/generate_azure_pipelines_matrices.py#L82-L83

and test commands here
https://github.com/miurahr/aqtinstall/blob/topic-refactor-azure/ci/steps.yml#L41-L46

miurahr:
I've merged master and resolve conflicts. Also change CI test code. Could you please pull from remote and see `ci/steps.yml`?

ddalcino:
Looks good; thanks @miurahr! I fixed some errors I saw on GH Actions, and added a couple more azure pipelines tests.

... oops, looks like I got one of the tests wrong ...

miurahr:
It looks good enough to merge, except for one test  expectation. Thanks @ddalcino for your hard  work !!🥇 

Could you add your name  in `docs/authors.rst` as a last task ?

miurahr:
Is it ready for review? If so. could you please change status by button?

image

ddalcino:
Sorry I'm slow with this; I've had some other things to do.

This is not quite done; I would like to reorganize a little before I mark it ready for review. I think I have added too many new functions to `helper.py`; they really should be wrapped up in their own class. I tried my best to make them potentially reusable by other code, but I don't think many of them will be very useful to anything but the `aqt list` command.

ddalcino:
This could definitely use some more tests. 

The `aqt list tools` feature is very unfinished. It needs to be able to list modules, architectures, etc, but I haven't implemented that yet. I thought I would save that part for another PR.

aqt/archives.py
```diff
@@ -37,6 +51,294 @@ def __init__(self, version, target, arch, os_name):
...
+        """
+        Construct ListCommand.
+        @param filter_minor         When set, the ListCommand will filter out all versions of
+                                    Qt that don't match this minor version.
+        @param is_latest_version    When True, the ListCommand will find all versions of Qt
+                                    matching filters, and only print the most recent version
+        @param modules_ver          Version of Qt for which to list modules
+        @param extensions_ver       Version of Qt for which to list extensions
+        @param architectures_ver    Version of Qt for which to list architectures
+        @param http_fetcher         Function used to fetch documents via http
```
miurahr, line 77: http_fetcher parameter seems to be used only for test purpose, not from Cli class.

I like  a way to override something ListCommand.fetcher() method using pytest monkeypatch.

miurahr (changes requested):
Many part looks good but  I'd like to ask changes for dependency and http_fetcher parameter.

aqt/helper.py
```diff
@@ -20,20 +20,145 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import configparser
+import dataclasses
```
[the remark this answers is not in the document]
ddalcino: Ok; I can get rid of this. This was a convenience, not absolutely necessary.

I had expected that since this was passing the CI tests that dataclasses were OK.

aqt/archives.py
ddalcino: This is correct; I am not accustomed to monkeypatching. This will be easy to change.

aqt/helper.py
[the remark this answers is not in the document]
miurahr: Current CI run only on 3.8 and 3.9, not 3.6  to reduce load for CI platform and mirror sites. sorry for inconvenience.

aqt/helper.py
```diff
@@ -188,6 +313,103 @@ def getlistint(self, section: str, option: str, fallback=[]):
...
+    """Make an HTTP request, using one or more base urls in case the request fails.
+    If all requests fail, then re-raise the requests.exceptions.RequestException
+    that was raised by the final HTTP request.
+    Any HTTP request that resulted in a status code >= 400 will result in a RequestException.
+    """
+    for i, base_url in enumerate(base_urls):
+        try:
+            url = base_url + "/" + rest_of_url
+            r = requests.get(url, timeout=timeout)
+            r.raise_for_status()
```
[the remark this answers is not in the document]
ddalcino: I'm not quite sure what you're asking for here. `requests.get` calls `requests.request`, documented [here](https://docs.python-requests.org/en/latest/api/#requests.request). By default, this function allows redirects (see the `allow_redirects` parameter). By the time the `requests.get` call returns, you should not ever see a redirect status code; you'll just see the response after redirection.

Regarding the timeout, every call to the `helper.request_http_with_failover` function comes from the default 'fetch http' function in ListCommand, which populates timeout with the value from Settings. I can add a default value that pulls in the Settings value directly, if that's what you're asking.

aqt/helper.py
```diff
@@ -25,15 +25,133 @@
...
+    def __str__(self):
+        return " ".join(self.strings)
+
+    def __bool__(self):
+        return len(self.strings) > 0 and len(self.strings[0]) > 0
+
+
+class Tools(ListOfStr):
+    def __str__(self):
+        return "\n".join(self.strings)
```
ddalcino, line 154: All of this could probably be moved directly into ListCommand.

docs/cli.rst
```diff
@@ -15,13 +15,81 @@ Generic commands
...
+
+.. option:: --tool <tool name>
+
+    The name of a tool. Use `aqt list tools <host> <target>` to see accepted values.
+    This flag only works with the 'tools' category, and may noy be combined with
+    any other flags.
+    When set, this prints all 'tool variant names' available.
+
+    The output of this command is meant to be used with the `aqt tool` command:
+    See the :ref:`Tool installation commands` below.
```
ddalcino, line 92: CI logs make it look like this will be a broken link. There's probably a setting somewhere to turn this feature on; I could try to figure it out if you want this link to work. Otherwise, I can just take the link out; I don't think the link is all that helpful.

aqt/helper.py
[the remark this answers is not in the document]
ddalcino: Again, I will need some guidance as to how to proceed here. 

I think it would make a lot of sense to get rid of the raw `request.get` here, and just use `helper.getUrl`, but i would need to make `helper.getUrl` more generic so that I can set the error messages properly. If a user tries to list modules for some Qt version that doesn't exist, I don't think it would make any sense to print "Download error!" when the user isn't trying to download anything. I would make those changes in a separate PR, because they're outside the scope of this feature.

docs/cli.rst
miurahr: We can add `extensions = ['sphinx.ext.autosectionlabel']` in docs/conf.py

docs/cli.rst
```diff
@@ -168,6 +237,13 @@ Example: Install an Install FrameWork (IFW):
     aqt tool linux tools_ifw 4.0 qt.tools.ifw.40
 
 
+Example: List the variants of IFW available:
+
+.. code-block:: bash
+
+    aqt list tools linux desktop --tool tools_ifw   # prints 'qt.tools.ifw.40'
```
miurahr, line 244: It is better to add more examples.
There are so many users who want to know  an instant solution for their needs.

docs/cli.rst
miurahr: I've pushed it. Please see.

miurahr:
I've pushed some additions to document. @ddalcino could you pull it and check it?

aqt/archives.py
```diff
@@ -38,6 +49,358 @@ def __init__(self, version, target, arch, os_name):
...
+
+    # Inner helper classes
+    class Versions:
+        def __init__(self, it_of_it: Iterable[Tuple[int, Iterable[Version]]]):
+            self.versions: List[List[Version]] = [
+                list(versions_iterator) for _, versions_iterator in it_of_it
+            ]
+
+        def __str__(self):
+            return "\n".join(
```
miurahr, line 63: IMHO, it is not good habit to include "\n" in __str__

aqt/archives.py
```diff
@@ -38,6 +49,358 @@ def __init__(self, version, target, arch, os_name):
...
+    def fetch_http(rest_of_url: str) -> str:
+        base_urls = Settings.baseurl, random.choice(Settings.fallbacks)
+        for i, base_url in enumerate(base_urls):
+            try:
+                url = posixpath.join(base_url, rest_of_url)
+                return getUrl(
+                    url=url,
+                    timeout=(Settings.connection_timeout, Settings.response_timeout),
+                    logger=getLogger("aqt"),
+                )
```
miurahr, line 268: I've changed logger handling in master branch.
We don't pass logger as argument, but get it on each class.

miurahr (changes requested):
I'd like to put `getLogger()` on function not caller.

aqt/archives.py
miurahr: This part is missing from the refactoring, sorry.

Please see. https://github.com/miurahr/aqtinstall/pull/294

ddalcino: Ok, this should be fixed now.

aqt/archives.py
ddalcino: Ok, I can change this.

While we're on this subject, maybe we should discuss exactly how the output for this feature should be formatted.

This feature prints different kinds of data in different ways:

* Versions of qt: Currently, we print all qt versions in ascending order, separated by spaces, with each patch number on a different line. For example:

      6.0.0 6.0.1 6.0.2 6.0.3 6.0.4
      6.1.0 6.1.1 6.1.2
      6.2.0

* Architectures, modules, extensions, and targets: These are all printed on a single line, with each item separated by a space character.
* Tools: These are all printed on separate lines. At first, I did this because many of the tool names are quite long, and it was easier for me to read this way.  In retrospect, I think that it's very strange that the tools aren't all printed on the same line like everything else. Also, it's strange that all of the tools begin with the prefix `tools_`. As a user, I would expect `aqt list tools` to print `ifw cmake qtcreator` instead of `tools_ifw tools_cmake tools_qtcreator`. It's a waste of horizontal space.

Is there anything on this list that you would like changed?

miurahr: I think `__str__` for human-readability, and pretty printing for CLI are different.
When logging object state to a log file in future, we will use `str()` in style `logger.log(obj)` that produce a line feed on single entry in a log file. It is not good.

Is it difficult define a named function for purpose of pretty print, and use it from CLI?

ddalcino: Sorry I did not make it more clear in my last message, but I intend to add a ‘pretty_print’ -type function and use it instead of ‘str’. That is easy to do. I just wanted to give you an opportunity to change what that ‘pretty_print’ function prints.

miurahr: It is ok as same as current `__str__()` doing.

> Tools: These are all printed on separate lines. At first, I did this because many of the tool names are quite long, and it was easier for me to read this way. In retrospect, I think that it's very strange that the tools aren't all printed on the same line like everything else. Also, it's strange that all of the tools begin with the prefix tools_. As a user, I would expect aqt list tools to print ifw cmake qtcreator instead of tools_ifw tools_cmake tools_qtcreator. It's a waste of horizontal space.

I agree with it. It is ok for me about current output.

ddalcino: Ok, it should be fixed now.

ddalcino:
I think it would make sense to squash a lot of this into fewer commits. I don't think you really want to add 85 commits to the master branch. Let me know what you think.

miurahr:
I'd like to squash and merge it.

miurahr:
Thank you very much!

PR #247, opened 2021-05-23
Add black to ci

ddalcino:
This causes the "Check Package" workflow to fail if the code hasn't been formatted before being pushed to Github. Here's an example of such a failure, after adding the check: https://github.com/ddalcino/aqtinstall/runs/2651537889?check_suite_focus=true.

It looks like four source files weren't formatted with black. The second commit in this PR formats those files, so the workflow doesn't fail.


miurahr:
The project has already use [flake8-black](https://pypi.org/project/flake8-black/)  to check code format with black rule.
It is why we don't need to run black individually in CI.

https://github.com/miurahr/aqtinstall/blob/d5f0bd6d411432b3d6edf4c369147934f76e1db0/pyproject.toml#L8-L10
https://github.com/miurahr/aqtinstall/blob/d5f0bd6d411432b3d6edf4c369147934f76e1db0/setup.cfg#L62
https://github.com/miurahr/aqtinstall/blob/d5f0bd6d411432b3d6edf4c369147934f76e1db0/tox.ini#L12

miurahr:
It seems [your change](https://github.com/miurahr/aqtinstall/commit/d16c9e1bcf54b991cef312db937bb93aba7f22db)  break running check.

The project confitures  tox  controlled by [tox-gh-actions](https://pypi.org/project/tox-gh-actions/)

https://github.com/miurahr/aqtinstall/blob/f61005c7c2d068a314fe8aa1df4754098b29d9c6/tox.ini#L41-L44

miurahr:
I've pushed PR #248 to revert two commits to fix check workflow, and fix PEP8/Black code formatting.

ddalcino:
Ok, I didn't realize that. Sorry!

Issue #256, opened 2021-06-03
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""


tsteven4:
My application fails with:

> QFactoryLoader::QFactoryLoader() checking directory path "/home/tsteven4/work/pvs2/gui/GPSBabelFE/platforms" ...
> qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
> This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
> 
> Aborted (core dumped)

This can be avoided by

1. setting QT_PLUGIN_PATH, e.g. /tmp/Qt/5.12.11/gcc_64/plugins
2. creating a link  /home/qt/work/install -> the installation, e.g. /tmp/Qt/5.12.11/gcc_64

I speculate that this occurs because the qmake binary is not patched.  An official install would have changed these strings in qmake to point to the installation:
qt_prfxpath=/opt/Qt5.12.10/5.12.10/gcc_64
qt_epfxpath=/opt/Qt5.12.10/5.12.10/gcc_64
qt_hpfxpath=/opt/Qt5.12.10/5.12.10/gcc_64

tsteven4:
My speculation is insufficient.  It appears the official installer patches up numerous other strings in other files as well.

It looks like a critical patch is qt_prfxpath in libQt5Core.so.5.12.11

miurahr:
Here is a source place about patching for qmake. An aqt does patch to qmake and QtCore.
 #100 is the issue related. 

What's missing? 

https://github.com/miurahr/aqtinstall/blob/master/aqt/updater.py#L92
https://github.com/miurahr/aqtinstall/blob/master/aqt/updater.py#L127

miurahr:
https://github.com/miurahr/aqtinstall/releases/tag/v1.0.0
aqtinstall v1.0.0 and later has a patch feature.

tsteven4:
Thanks, I seem to  be downlevel and I don't know how to get a more 
recent version.

 From aqtinstall/README.rst at master · miurahr/aqtinstall · GitHub 
<https://github.com/miurahr/aqtinstall/blob/master/README.rst>

> $ pip install -U pip
> $ pip install aqtinstall
got me

> pip show aqtinstall
> Name: aqtinstall
> Version: 0.11.1

miurahr:
```
$ pip install --help

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...

[...]

Issue #100, opened 2020-02-27
qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""


MartinDelille:
I installed Qt 5.12.7 with `aqt install 5.12.7 mac desktop` and installed *Qt Creator* via [Homebrew](https://brew.sh) (`brew cask install qt-creator`).

I configured *Qt Creator* kits to use the one install via `aqt` and I can compile a simple widget project.

Unfortunately, I have the following error when starting it in debug mode:

```
QFactoryLoader::QFactoryLoader() checking directory path "/Users/martin/dev/build/pouet/Qt_5_12_7_clang_64_from_aqt/Debug/pouet.app/Contents/MacOS/platforms" ...
qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

The program has unexpectedly finished.
```

`QT_DEBUG_PLUGINS` is set to `1` as adviced here: https://forum.qt.io/topic/98816/qt-could-not-find-the-platform-plugin-cocoa/8

Any idea what's going on?

miurahr:
It seems installed at "5.12.7/clang_64/plugins/platforms/' please check in your installation.
The package seems https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt5_5127/qt.qt5.5127.clang_64/5.12.7-0-202001290638qtbase-MacOS-MacOS_10_13-Clang-MacOS-MacOS_10_13-X86_64.7z

MartinDelille:
It seems that the `platforms` directory is ok:

```
$ls /path/to/aqt/5.12.7/clang_64/plugins/platforms/
libqcocoa.dylib
libqcocoa_debug.dylib
libqminimal.dylib
libqminimal_debug.dylib
libqoffscreen.dylib
libqoffscreen_debug.dylib
```

MartinDelille:
Ok sorry the problem doesn't come from `aqt` but from the *Qt Creator* version installed with *Homebrew cask*

MartinDelille:
Ok I tried with an official installer downloaded from The Qt Company and compiled and run a basic widget project with two kit:

1. The one configured automatically (5.12.6)
2. The one configured manually by selecting the `qmake` binary install with `aqt`

I only have the error in the second case.

I'm really interested in `aqt` so I'll be happy to dig more into this problem. Could you reopen the issue?

[...]

PR #257, opened 2021-06-05
fix binfile patching.

tsteven4:
I think this is the fix for #256.  Be aware I am unfamiliar with any test process you may have, and I do not regularly program in python.

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -36,10 +36,10 @@ class Updater:
         st = file.stat()
         data = file.read_bytes()
         idx = data.find(key)
-        if idx > 0:
+        if idx < 0:
             return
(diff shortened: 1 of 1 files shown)


miurahr:
Thank you for finding an ugly problem in patching code. @tsteven4  great work!🚀 

Patching binary was introduced in v0.9.0 in May, 2020 by #130 so the bug  exist in all v0.9.0  - v1.1.6

miurahr:
A patching qmake was discussed in #100 and refers to a code snippet   https://gist.github.com/agateau/f217cebd8b645b1c7466d233e1f1fcfb and use it in #130.

miurahr:
@tsteven4 I'd like to add you name to `docs/authros.rst`. Could you give a name to be added?

PR #258, opened 2021-06-05
make sure binfile patches are a full 256 bytes, including

tsteven4:
any necessary padding.

This also fixes the assertion regarding the length of the new path.

This is an additional required fix for #256.  The need for this additional fix was pointed out in #256 by @ddalcino.

Again, please check my python.  The intent is that every patch insertion shall be exactly 256 bytes consisting of 
1) the key
2) the new path
3) zero padding of at least 1 byte

It's not clear to me that this fixed length string needs to be null terminated, but it might, and it the reduction of the maximum new path length by 1 byte seems an insignificant concession.

I believe python assertions only run with __debug__ is true, which only happens when optimization is not requested. https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement
Therefore we may want to further change the length check so it always runs.

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -39,12 +39,13 @@ class Updater:
         if idx < 0:
             return
         assert len(newpath) < 256, "Qt Prefix path is too long(255)."
-        data = (
-            data[: idx + len(key)]
-            + newpath
-            + bytes(256 - len(newpath))
-            + data[idx + len(key) + 256 :]
-        )
+        oldlen = data[idx + len(key) :].find(b"\0")
+        assert oldlen >= 0
+        if oldlen > len(newpath):
+            value = newpath + bytes(oldlen - len(newpath))
+        else:
+            value = newpath
+        data = data[: idx + len(key)] + value + data[idx + len(key) + len(value) :]
         file.write_bytes(data)
         os.chmod(str(file), st.st_mode)
 
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -41,10 +41,7 @@ class Updater:
         assert len(newpath) < 256, "Qt Prefix path is too long(255)."
         oldlen = data[idx + len(key) :].find(b"\0")
         assert oldlen >= 0
-        if oldlen > len(newpath):
-            value = newpath + bytes(oldlen - len(newpath))
-        else:
-            value = newpath
+        value = newpath + b"\0" * (oldlen - len(newpath))
         data = data[: idx + len(key)] + value + data[idx + len(key) + len(value) :]
         file.write_bytes(data)
         os.chmod(str(file), st.st_mode)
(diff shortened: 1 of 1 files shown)


ddalcino:
@miurahr's comment in #256 shows this code, which appears to be the definition of the key-value pairs in the code that builds qmake:

```
$${currentConfig}.output.qconfigSource = \
    "/* Installation Info */" \
    "static const char qt_configure_prefix_path_str  [12+256] = \"qt_prfxpath=$$config.input.prefix\";" \
    "$${LITERAL_HASH}ifdef QT_BUILD_QMAKE" \
    "static const char qt_configure_ext_prefix_path_str   [12+256] = \"qt_epfxpath=$$config.input.extprefix\";" \
    "static const char qt_configure_host_prefix_path_str  [12+256] = \"qt_hpfxpath=$$config.input.hostprefix\";" \
    "$${LITERAL_HASH}endif" \
    "" \
```

Each key-value pair takes up 12 bytes for the `key=` part and 256 bytes for the value, which appears to be null terminated. 

This PR would shorten the space for the value to 244 bytes, or 243 bytes without the null terminal. I don't think that's really what you want.

tsteven4:
Another good catch, I assumed they used 256 byte blocks.  From the dump it looks like they use 288 byte blocks even though the pri file shows 268 byte blocks.  Safest to go with what the pri indicates and leave the other 20 bytes alone.

The pri does raise the question of the need to patch the other two  paths, qt_epfxpath and qt_hpfxpath.

I think this all becomes mute for Qt >= 5.14, when Qt installs are suppose to be reolcatable. https://www.qt.io/blog/qt-is-relocatable.

ddalcino:
Thanks.

I suspect that the compiler is aligning these strings to 32 byte blocks. Choice of alignment size is usually a function of compiler/host architecture/OS; it would definitely not be safe to claim the last 20 bytes on some other platform where the compiler chooses a different alignment.

Good point re: `qt_epfxpath` and `qt_hpfxpath`. The `Updater._patch_binfile` function looks generic enough to be able to patch both of these, but it is only called for `qt_prfxpath`.

On my copy of qmake, `qt_prfxpath`, `qt_epfxpath` and `qt_hpfxpath` are all set to the same value. I wonder what conditions would cause you to need to use a different value for the last two.

P.S.: the CI check is telling you to run `black` to auto-format your code before you commit.

tsteven4:
I changed based on the way the qt installer works: https://code.qt.io/cgit/installer-framework/installer-framework.git/tree/src/libs/installer/qtpatch.cpp?id=0ff7968345752395af93334b174973ebefb95048

It compares the lengths of the old and new value and only zeroes bytes in the old value if necessary.

miurahr:
thank you!

PR #259, opened 2021-06-05
Add test and control for patched qmake

ddalcino:
This adds two build jobs to the azure pipeline: one that tests a really short output directory, and one for a really long output directory.

This also directly tests the values `qt_prfxpath`, `qt_epfxpath`, and `qt_hpfxpath` in the qmake binary, in the two new build jobs (perhaps this check should be extended to all the build jobs?)

This also hides `<qt_version>/<arch>/bin/qt.conf` in the two new build jobs, to verify that the qmake binary has been patched properly.

This also turns on `xtrace` in `ci/steps.yml`; this made it a lot easier for me to debug the terminal output in the Azure Pipelines jobs. This change probably belongs in a separate PR; please let me know if you'd like me to split this up.

### DIFF
--- a/ci/steps.yml
+++ b/ci/steps.yml
@@ -12,7 +12,7 @@ steps:
   ## we insert sleep in random duration < 30sec to reduce
   ## download server load.
   - bash: |
-      set -e
+      set -ex
       number=$RANDOM
       let "number %= 30"
       sleep $number
@@ -26,6 +26,11 @@ steps:
         if [[ "$(MODULE)" != "" ]]; then
           opt+=" -m $(MODULE)"
         fi
+        if [[ "$(OUTPUT_DIR)" != "" ]]; then
+          opt+=" --outputdir $(OUTPUT_DIR)"
+          sudo mkdir -p "$(OUTPUT_DIR)"
+          sudo chown $(whoami) "$(OUTPUT_DIR)"
+        fi
         if [[ "$(SUBARCHIVES)" != "" ]]; then
           opt+=" --archives $(SUBARCHIVES)"
         fi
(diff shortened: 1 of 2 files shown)


ci/steps.yml
```diff
@@ -37,6 +42,26 @@ steps:
...
+          echo "Hide qt.conf so it doesn't interfere with test"
+          mv $(QT_BINDIR)/qt.conf $(QT_BINDIR)/_qt.conf
+          #export PATH=$(QT_BINDIR):$PATH
+          echo "Require that qt_epfxpath was set to '$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)'"
+          [[ $($(QT_BINDIR)/qmake -query QT_INSTALL_PREFIX) == "$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)" ]]
+          echo "Require that qt_prfxpath was set to '$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)'"
+          [[ $($(QT_BINDIR)/qmake -query QT_INSTALL_PREFIX/dev) == "$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)" ]]
+          echo "Require that qt_hpfxpath was set to '$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)'"
+          [[ $($(QT_BINDIR)/qmake -query QT_HOST_PREFIX) == "$(OUTPUT_DIR)/$(QT_VERSION)/$(ARCHDIR)" ]]
+        fi
```
ddalcino, line 64: These checks are redundant; the three checks using `strings` above here are sufficient. I have added both with the expectation that we would remove either the `strings` check or the `qmake -query` check. There's no good reason to keep both.

ci/steps.yml
```diff
@@ -67,7 +92,7 @@ steps:
       export ANDROID_NDK_ROOT=$(Build.SourcesDirectory)/android-ndk-r21e
       mkdir $(Build.BinariesDirectory)/tests
       (cd $(Build.BinariesDirectory)/tests; 7zr x $(Build.SourcesDirectory)/ci/accelbubble.7z)
-      export PATH=$(Build.BinariesDirectory)/Qt/$(QT_VERSION)/$(ARCHDIR)/bin:$PATH
+      export PATH=$(QT_BINDIR):$PATH
```
ddalcino, line 95: This is part of a global replacement of `$(Build.BinariesDirectory)/Qt/$(QT_VERSION)/$(ARCHDIR)/bin` for `$(QT_BINDIR). I found it simpler to use a variable than to conditionally change the PATH variable.

ci/steps.yml
```diff
@@ -139,7 +165,7 @@ steps:
           python -m aqt tool --outputdir $(Build.BinariesDirectory)/Qt $(HOST) tools_mingw 8.1.0-1-202004170606 qt.tools.win32_mingw810
           [Environment]::SetEnvironmentVariable("Path", ";$(Build.BinariesDirectory)\Qt\Tools\mingw810_32\bin" + $env:Path, "Machine")
         }
-        $env:Path = "$(Build.BinariesDirectory)\Qt\Tools\$(ARCHDIR)\bin;$(Build.BinariesDirectory)\Qt\$(QT_VERSION)\$(ARCHDIR)\bin;" + $env:Path
+        $env:Path = "$(Build.BinariesDirectory)\Qt\Tools\$(ARCHDIR)\bin;$(WIN_QT_BINDIR);" + $env:Path
```
ddalcino, line 168: Similar to `QT_BINDIR`, I made a `WIN_QT_BINDIR` to account for the different path separator on Windows. I'm not sure how well Powershell handles paths with forward slashes.

We could get away with not using `WIN_QT_BINDIR` here if we're not adding these tests to any of the Powershell builds. I'm adding this as a precaution in case someone wants to add such a test in the future. Whoever does that will need to either conditionally replace `$(Build.BinariesDirectory)\Qt\Tools\$(ARCHDIR)\bin` with `$(OUTPUT_DIR)\Tools\$(ARCHDIR)\bin`, or add something like a `WIN_QT_TOOLS_BINDIR` variable.

ci/steps.yml
```diff
@@ -12,7 +12,7 @@ steps:
   ## we insert sleep in random duration < 30sec to reduce
   ## download server load.
   - bash: |
-      set -e
+      set -ex
```
ddalcino, line 15: This turns on `xtrace`, which prints the values of all the variable substitutions in the Azure Pipelines logs. I couldn't figure out any other way to debug this file other than to do this. I can go back and take these out if you prefer.

PR #263, opened 2021-06-11
Version comparison by semantic_version

miurahr:


aqt/archives.py
```diff
@@ -17,19 +17,24 @@
 # IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
 # FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
 # COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
-# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR I    packagingN
```
miurahr, line 20: typo

aqt/archives.py
```diff
@@ -76,26 +81,26 @@ def __init__(self, version, os_name, target, base, timeout=(5, 5)):
...
-        self.qt_ver_base = self.version[0:1]
         # Get packages index
-        if self.qt_ver_base == "6" and self.target == "android":
+        if self.version.major == 6 and self.target == "android":
             arch_ext = ["_armv7/", "_x86/", "_x86_64/", "_arm64_v8a/"]
         elif (
-            self.qt_ver_base == "5"
-            and int(qt_ver_num) >= 5130
+            self.version.major == 5
+            and self.version >= Version("5.13.0")
```
miurahr, line 89: Here can be `self.version in SimpleSpec(">=5.13.0,<6.0")`

aqt/archives.py
```diff
@@ -17,19 +17,24 @@
...
+
 from aqt.exceptions import ArchiveListError, NoPackageFound
 from aqt.helper import Settings, getUrl
 
 
 class TargetConfig:
     def __init__(self, version, target, arch, os_name):
-        self.version = version
+        if isinstance(version, Version):
+            self.version = version.public
```
ddalcino, line 35: `version.public` is a data member of `packaging.version.Version`, but not `semantic_version.Version`.

If you need to turn a `Version` into a string, you can always use `str(version)`.  Both version classes provide a `.__str__()` method.

aqt/archives.py
ddalcino: Sounds like a good use of SimpleSpec to me.

aqt/archives.py
```diff
@@ -155,44 +160,69 @@ def __init__(
...
                 self.mod_list.append(
-                    "qt.qt{}.{}.{}.{}".format(self.qt_ver_base, qt_ver_num, m, arch)
+                    "qt.qt{0}.{1}{2}{3}.{4}.{5}".format(
+                        self.version.major,
+                        self.version.major,
+                        self.version.minor,
+                        self.version.patch,
+                        m,
+                        arch,
+                    )
```
ddalcino, line 175: There are a couple of these that could be a little less redundant. You could use:
```python
 "qt.qt{0}.{0}{1}{2}.{3}.{4}".format(
    self.version.major,
    self.version.minor,
    self.version.patch,
    m,
    arch,
)
```
...but that's a stylistic choice, and probably not very important.

aqt/archives.py
```diff
@@ -371,15 +408,15 @@ def _parse_update_xml(self, archive_url, target_packages):
...
                     downloadable_archives = []
-                full_version = packageupdate.find("Version").text
-                if not full_version.startswith(self.version):
+                named_version = packageupdate.find("Version").text
+                full_version = Version(named_version)
+                if not full_version.base_version == self.version.base_version:
                     self.logger.warning(
-                        "Version {} differ from requested version {} -- skip.".format(
-                            full_version, self.version
+                        "Base Version of {} is differed from requested version {} -- skip.".format(
```
ddalcino, line 415: I think you mean `is different` instead of `is differed`.

aqt/updater.py
```diff
@@ -230,18 +229,18 @@ def update(cls, target, base_dir: str, logger):
                 "android_x86",
                 "android_armv7",
             ]:  # desktop version
-                updater.make_qtconf(base_dir, qt_version, arch_dir)
+                updater.make_qtconf(base_dir, target.version, arch_dir)
                 updater.patch_qmake()
                 if target.os_name == "linux":
                     updater.patch_pkgconfig()
                 if versiontuple(target.version) < (5, 14, 0):
```
ddalcino, line 236: I would really like this line to look more like:
```python
if target.version < Version("5.14.0"):
```

aqt/archives.py
miurahr: yes.

aqt/updater.py
```diff
@@ -230,18 +229,18 @@ def update(cls, target, base_dir: str, logger):
...
                     updater.patch_qtcore(target)
-            elif qt_version.startswith("5."):  # qt5 non-desktop
+            elif Version(target.version) in SimpleSpec(">=5.0,<6.0"):
```
ddalcino, line 238: We appear to be missing a `from semantic_version import Version` at the top of the file here.

ddalcino:
I think it looks pretty good. It's definitely an improvement.

I have one other question though: In `aqt.archives.PackageList` and `aqt.archives.QtArchives`, you're storing datamembers called `version` that are `Version` objects. In `aqt.archives.TargetConfig`, you're storing a datamember called `version` that is a string object, not a Version. In several places where you're using `aqt.archives.TargetConfig` objects, you're converting the `version` string back into a `Version` object to do some comparisons. Is there a reason for this? I wonder if it would make more sense to store it as a Version object in the TargetConfig. I think it might be easier to read the code if every object with the word `version` in the name was actually a `Version` object, and harder to introduce new bugs.

Also, I think there's still an unused import: `aqt/installer.py:36:1: F401 'semantic_version.SimpleSpec' imported but unused`

miurahr:
> I have one other question though:

Current `aqt.archives.TargetConfig` can have `version`  as Qt version and other special one, such as "Tool", "Doc" etc. in same variable. It is why I hold version as str.

tsteven4:
While you are at it please add 5.12.11 to combinations.json to avoid
```
- aqt - WARNING - Specified Qt version is unknown: 5.12.11.
```

PR #267, opened 2021-06-12
patch pkgconfig files on mac, libtool control files on linux and mac

### LINKED ISSUE #265: pkgconfig files not patched on macOS
they exist in lib/pkgconfig, and are patched by the official installer.

tsteven4:
This resolves #265.

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -99,7 +99,7 @@ class Updater:
                 "-F{}".format(os.path.join(str(self.prefix), "lib")),
             )
 
-    def patch_libtool(self, oldvalue):
+    def patch_libtool(self, oldvalue, os_name):
         for lafile in self.prefix.joinpath("lib").glob("*.la"):
             self.logger.info("Patching {}".format(lafile))
             self._patch_textfile(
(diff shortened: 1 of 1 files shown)


aqt/updater.py
```diff
@@ -85,14 +85,24 @@ def _detect_qmake(self) -> bool:
...
+                self._patch_textfile(
+                    pcfile,
+                    "prefix=/home/qt/work/install",
+                    "prefix={}".format(str(self.prefix)),
+                )
+            elif os_name == "mac":
+                self.logger.info("Patching {}".format(pcfile))
+                self._patch_textfile(
+                    pcfile,
+                    "prefix=/Users/qt/work/install",
```
miurahr, line 101: There is only a difference is a value here.
I prefer to determine os_name  in update() function, and pass the value `oldprefix='/Users/qt/work/install'` as an argument of patch_pkgconfig function. It will reduce a code duplication.

tsteven4:
There are additional locations in the package config files that need patching as well.  And the lib/*.la files need patching as well.

aqt/updater.py
tsteven4: done.

miurahr:
Thank you.

miurahr:
v1.2.0b4 with the fix out.

PR #279, opened 2021-06-22
create qtenv2.bat file on windows

### LINKED ISSUE #278: on windows bin/qtenv2.bat should be created by installer

tsteven4:
This resolves #278

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -212,6 +212,23 @@ class Updater:
             f.write("[Paths]\n")
             f.write("Prefix=..\n")
 
+    def make_qtenv2(self, base_dir, qt_version, arch_dir):
+        """Prepare qtenv2.bat"""
+        with open(
+            os.path.join(base_dir, qt_version, arch_dir, "bin", "qtenv2.bat"), "w"
+        ) as f:
+            f.write("@echo off\n")
+            f.write("echo Setting up environment for Qt usage...\n")
+            f.write(
+                "set PATH={};%PATH%\n".format(
+                    os.path.join(base_dir, qt_version, arch_dir, "bin")
+                )
+            )
+            f.write("cd /D {}\n".format(os.path.join(base_dir, qt_version, arch_dir)))
+            f.write(
+                "echo Remember to call vcvarsall.bat to complete environment setup!\n"
+            )
+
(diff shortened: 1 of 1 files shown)


miurahr:
Thank you.
Check fails with another reasons.

pzhlkj6612:
Hi, @tsteven4 . I think qtenv2.bat may need to be modified to match the behavior of the batch file created by the official Qt online installer.

I have installed Qt for msvc2019_64 and mingw81_64 on my PC. There are two qtenv2.bat files in Qt installation directory, the contents of which are as follows.

msvc2019_64 (C:\Qt\5.15.2\msvc2019_64\bin\qtenv2.bat):

```bat
@echo off
echo Setting up environment for Qt usage...
set PATH=C:\Qt\5.15.2\msvc2019_64\bin;%PATH%
cd /D C:\Qt\5.15.2\msvc2019_64
echo Remember to call vcvarsall.bat to complete environment setup!

```

This is what this PR wants to do. I think it should work great. However,

mingw81_64 (C:\Qt\5.15.2\mingw81_64\bin\qtenv2.bat):

```bat
@echo off
echo Setting up environment for Qt usage...
set PATH=C:\Qt\5.15.2\mingw81_64\bin;C:/Qt/Tools/mingw810_64\bin;%PATH%
cd /D C:\Qt\5.15.2\mingw81_64

```

Please note the line including `C:/Qt/Tools/mingw810_64\bin` . MinGW is not part of Qt binary, but as a tool. The official Qt online installer (I'm using 4.1.1-0-202105251051) will automatically install the corresponding version of MinGW tool when the user wants to install Qt for MinGW, even if that MinGW tool is not checked (selected) in the tree of "Developer and Designer Tools". During the installation, a qtenv2.bat file will be created with the path of the installed MinGW tool.

Possible modifications:

1. If the user didn't install Qt for MinGW, qtenv2.bat should not say "Remember to call vcvarsall.bat to complete environment setup!" because it's only used for MSVC.
2. If the user installed Qt for MinGW, it should tell the user this file is not the same as the one created by the official Qt online installer, and the path to the "bin" directory of a valid and suitable MinGW should be added into "PATH" manually by the user.

Welcome to discuss.

PR #281, opened 2021-06-25
Fix wrong base path in run_list

### LINKED ISSUE #280: Wrong url in run_list
```
$ aqt list 5.15.0 linux desktop
aqtinstall(aqt) v1.2.1 on Python 3.8.3 [CPython GCC 7.5.0]
Download error when access to https://ftp.jaist.ac.jp/pub/qtprojectlinux_x64/desktop/qt5_5150/Updates.xml
```

Called from https://github.com/miurahr/aqtinstall/blob/67506c053b0dae40bc4f36fca1bc42965df868c8/aqt/installer.py#L465-L477

it's missing online/qtsdkrepository in the path.

As a side node, I'd say you should considering using `posixpath` to join urls to avoid having to deal with '/' being present or not.

jmarrec:
Fixed #280

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -20,6 +20,7 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import posixpath
 import xml.etree.ElementTree as ElementTree
 from logging import getLogger
 
@@ -72,7 +73,7 @@ class PackagesList:
         self.os_name = os_name
         self.target = target
         self.archives = []
-        self.base = base
+        self.base = posixpath.join(self.base, "online", "qtsdkrepository")
         self.timeout = timeout
         self.logger = getLogger("aqt")
         self._get_archives()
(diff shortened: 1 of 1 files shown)


miurahr:
Thank you! @jmarrec

aqt/archives.py
```diff
@@ -72,7 +73,7 @@ def __init__(self, version, os_name, target, base, timeout=(5, 5)):
         self.os_name = os_name
         self.target = target
         self.archives = []
-        self.base = base
+        self.base = posixpath.join(self.base, "online", "qtsdkrepository")
```
ddalcino, line 76: This change causes the parameter `base` to be ignored. Was this line meant to be:

```python
self.base = posixpath.join(base, "online", "qtsdkrepository")
```

When I try to run `aqt list  5.15.2 linux desktop` after this change, I get a stack trace ending in:

```bash
  File "D:\Documents\PycharmProjects\aqtinstall\aqt\archives.py", line 76, in __init__
    self.base = posixpath.join(self.base, "online", "qtsdkrepository")
AttributeError: 'PackagesList' object has no attribute 'base'
```

jmarrec: You're absolutely right, sorry!

jmarrec: https://github.com/miurahr/aqtinstall/pull/286/

Issue #260, opened 2021-06-09
bad version comparison


tsteven4:
The compare at L86 will not give the desired result for some versions, e.g. 5.12.10.  That is 51210 is >= 5130.

https://github.com/miurahr/aqtinstall/blob/bffc4b81a7c528779a42a082306274523f65ff9b/aqt/archives.py#L78-L86

As a result the list command can fail:
```
tsteven4@focal:~/.local/lib/python3.8/site-packages$ aqt list 5.12.10 linux desktop
2021-06-09 09:44:21,765 - aqt - INFO - aqtinstall(aqt) vunknown on Python 3.8.5 [CPython GCC 9.3.0]
Traceback (most recent call last):
  File "/home/tsteven4/.local/bin/aqt", line 8, in <module>
    sys.exit(main())
  File "/home/tsteven4/.local/lib/python3.8/site-packages/aqt/__init__.py", line 38, in main
    return cli.run()
  File "/home/tsteven4/.local/lib/python3.8/site-packages/aqt/installer.py", line 675, in run
    return args.func(args)
  File "/home/tsteven4/.local/lib/python3.8/site-packages/aqt/installer.py", line 482, in run_list
    pl = PackagesList(qt_version, host, target, BASE_URL)
  File "/home/tsteven4/.local/lib/python3.8/site-packages/aqt/archives.py", line 92, in __init__
    self._get_archives()
  File "/home/tsteven4/.local/lib/python3.8/site-packages/aqt/archives.py", line 119, in _get_archives
    self.update_xml = ElementTree.fromstring(r.text)
  File "/usr/lib/python3.8/xml/etree/ElementTree.py", line 1320, in XML
    parser.feed(text)
xml.etree.ElementTree.ParseError: syntax error: line 1, column 49
```
In addition, 5.12.11 needs to be added to combinations.json.

ddalcino:
I noticed this earlier as well. Thank you for making this issue, I probably would have forgotten to do so.

This comparison cannot be done safely as an integer; I think it should be done using a Version class that supports comparisons. I wound up using Version objects from [the semantic-version library](https://pypi.org/project/semantic-version/) in #235, but I didn't realize that the project already uses the Version class in [the packaging library](https://packaging.pypa.io/en/latest/version.html), in [`Cli._set_arch()`](https://github.com/miurahr/aqtinstall/blob/master/aqt/installer.py#L147). I don't know which library is better for this purpose.

[...]

PR #283, opened 2021-06-25
Fix #282 - Refactor archives.py and make `aqt tool` take --version and --arch as optionals

### LINKED ISSUE #282: Enhancement Request: Change `aqt tool` to not require version and arch (--latest)
I can see that `tools_ifw` always breaks because Qt will remove any previous versions (there's only one). It seems we could just parse the Updates.xml and get the latest one if the version isn't actually supplied.

Also, it seems like error reporting needs to be improved. Even if the tool isn't found, the "Installation finished" is issued.

I will try to provide a PR for review

jmarrec:
- Use packaging.version instead of semantic_version
    - That way you could pass --version 4, it would allow building a version, and it would match any version that starts with 4. See the helper I added, and the unit tests that show the behavior clearly  
- Provide `__str__`, `__repr__` for classes for easier debugging
- Use `posixpath` instead of string manipulation to make urls
- Add type hints were I touched thing
   -  TODO: should add mypi as a checker

Edit: it's likely that some of the workflows will fail, and that's fine. I only ran the unit test locally + try `aqt tool`, haven't tried Qt. We can address problems as they are caught

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -20,13 +20,14 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import posixpath
 import xml.etree.ElementTree as ElementTree
 from logging import getLogger
+from typing import Optional, Tuple
+
 import packaging.version
 from packaging.version import Version
-import posixpath
 
-from typing import Optional, Tuple
 from aqt.exceptions import ArchiveListError, NoPackageFound
 from aqt.helper import Settings, getUrl, satifiesVersion
 
@@ -39,11 +40,13 @@ class TargetConfig:
         self.os_name = os_name
 
     def __str__(self):
-        print(f'TargetConfig(version={self.version}, target={self.target}, '
-              f'arch={self.arch}, os_name={self.os_name}')
+        print(
+            f"TargetConfig(version={self.version}, target={self.target}, "
+            f"arch={self.arch}, os_name={self.os_name}"
+        )
 
     def __repr__(self):
-        print(f'({self.version}, {self.target}, {self.arch}, {self.os_name})')
+        print(f"({self.version}, {self.target}, {self.arch}, {self.os_name})")
 
 
 class QtPackage:
@@ -51,9 +54,15 @@ class QtPackage:
     Hold package information.
     """
 
-    def __init__(self, name: str , archive_url: str, archive: str,
-                 package_desc: str, hashurl: str,
-                 version: Optional[Version] = None):
+    def __init__(
+        self,
+        name: str,
+        archive_url: str,
+        archive: str,
+        package_desc: str,
+        hashurl: str,
+        version: Optional[Version] = None,
+    ):
         self.name = name
         self.url = archive_url
         self.archive = archive
@@ -62,16 +71,16 @@ class QtPackage:
         self.version = version
 
     def __repr__(self):
-        v_info = f', version={self.version}' if self.version else ''
-        return (
-            f'QtPackage(name={self.name}, archive={self.archive}{v_info})'
-        )
+        v_info = f", version={self.version}" if self.version else ""
+        return f"QtPackage(name={self.name}, archive={self.archive}{v_info})"
 
     def __str__(self):
-        v_info = f', version={self.version}' if self.version else ''
-        return (f'QtPackage(name={self.name}, url={self.url}, '
-                f'archive={self.archive}, desc={self.desc}'
-                f'hashurl={self.hashurl}{v_info})')
+        v_info = f", version={self.version}" if self.version else ""
+        return (
+            f"QtPackage(name={self.name}, url={self.url}, "
+            f"archive={self.archive}, desc={self.desc}"
+            f"hashurl={self.hashurl}{v_info})"
+        )
 
 
 class ListInfo:
@@ -106,9 +115,11 @@ class PackagesList:
         # Get packages index
         if self.version.major == 6 and self.target == "android":
             arch_ext = ["_armv7/", "_x86/", "_x86_64/", "_arm64_v8a/"]
-        elif ((self.target == 'desktop') and
-              (self.version >= Version('5.13.0')) and
-              (self.version < Version('6.0.0'))):
+        elif (
+            (self.target == "desktop")
+            and (self.version >= Version("5.13.0"))
+            and (self.version < Version("6.0.0"))
+        ):
             arch_ext = ["/", "_wasm/"]
         else:
             arch_ext = ["/"]
@@ -165,7 +176,9 @@ class QtArchives:
         timeout=(5, 5),
     ):
         self.version_str = version_str
-        self.version = packaging.version.parse(version_str) if version_str is not None else None
+        self.version = (
+            packaging.version.parse(version_str) if version_str is not None else None
+        )
         self.target = target
         self.arch = arch
         self.os_name = os_name
@@ -237,9 +250,9 @@ class QtArchives:
             )
         )
         target_packages.extend(self.mod_list)
-        self.logger.debug('QtArchives::_get_archives')
-        self.logger.debug(f'{update_xml_url=}')
-        self.logger.debug(f'{target_packages=}')
+        self.logger.debug("QtArchives::_get_archives")
+        self.logger.debug(f"{update_xml_url=}")
+        self.logger.debug(f"{target_packages=}")
 
         self._download_update_xml(update_xml_url)
         self._parse_update_xml(archive_url, target_packages)
@@ -277,7 +290,7 @@ class QtArchives:
                     ).text.split(", ")
                     full_version = packageupdate.find("Version").text
                     # keep only '5.15.2' in '5.15.2-0-202011130724'
-                    parsed_version = packaging.version.parse(full_version.split('-')[0])
+                    parsed_version = packaging.version.parse(full_version.split("-")[0])
                     package_desc = packageupdate.find("Description").text
                     for archive in downloadable_archives:
                         archive_name = archive.split("-", maxsplit=1)[0]
@@ -287,7 +300,7 @@ class QtArchives:
                             # qt.qt5.5150.gcc_64/
                             name,
                             # 5.15.0-0-202005140804qtbase-Linux-RHEL_7_6-GCC-Linux-RHEL_7_6-X86_64.7z
-                            full_version + archive
+                            full_version + archive,
                         )
                         hashurl = package_url + ".sha1"
                         self.archives.append(
@@ -297,7 +310,7 @@ class QtArchives:
                                 archive,
                                 package_desc,
                                 hashurl,
-                                parsed_version
+                                parsed_version,
                             )
                         )
         if len(self.archives) == 0:
@@ -305,7 +318,7 @@ class QtArchives:
                 "Specified packages are not found while parsing XML of package information!"
             )
             raise NoPackageFound
-        self.logger.debug(f'Archives: {self.archives}')
+        self.logger.debug(f"Archives: {self.archives}")
 
     def get_archives(self):
         """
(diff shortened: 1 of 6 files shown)


aqt/helper.py
```diff
@@ -205,6 +207,17 @@ def handle(self, record):
...
 
+def satifiesVersion(requestedVersion: Version, candidateVersion: Version):
+    """
+    This will return true if the candidateVersion matches the requestedVersion
+    It will only compare the number of components of the candidateVersion
+    """
+    for (a, b) in zip(requestedVersion.release, candidateVersion.release):
+        if a != b:
+            return False
+    return True
```
jmarrec, line 218: New helper.

v.release returns a tuple

```
v1 = Version('4.0') => v1.release = (4, 0)
v2 = Version('4.0.1') => v2.release = (4, 0, 1)

zip:
  4 == 4
  0 == 0
  
  => it matches
```

aqt/installer.py
```diff
@@ -427,13 +427,14 @@ def run_tool(self, args):
...
-                os_name,
-                tool_name,
-                version,
-                arch,
-                base,
+                os_name=os_name,
+                tool_name=tool_name,
+                base=base,
+                version_str=version,
+                arch=arch,
```
jmarrec, line 437: Avoid positional args, so we can reorder without problems

aqt/installer.py
```diff
@@ -635,10 +636,17 @@ def _create_parser(self):
...
+            type=str,
+            help=('Tool version in the format of "4.1.2". '
+                  'If not supplied, defaults to latest found'),
         )
+
         tools_parser.add_argument(
-            "arch", help="Name of full tool name such as qt.tools.ifw.31"
+            '-a', '--arch',
+            type=str,
+            help='Name of full tool name such as "qt.tools.ifw.31"'
```
jmarrec, line 649: Make these two optionals

aqt/installer.py
```diff
@@ -676,6 +684,8 @@ def run(self, arg=None):
         return result
 
     def call_installer(self, qt_archives, base_dir, sevenzip, keep):
+        if not qt_archives.get_archives():
+            raise NoPackageFound("Couldn't find any archives")
```
jmarrec, line 688: I will be raising earlier in `ToolsArchives` with a better error message, but I think it's fair to have this as a fallback...

tests/test_helper.py
```diff
@@ -100,3 +102,26 @@ def test_helper_downloadBinary_sha256(tmp_path, monkeypatch):
...
+    ],
+)
+def test_satifiesVersion(requestedVersion, candidateVersion, expected):
+    v1 = packaging.version.parse(requestedVersion)
+    v2 = packaging.version.parse(candidateVersion)
+    matches = helper.satifiesVersion(v1, v2)
+    if expected:
+        assert matches, f"Expected {candidateVersion} to satify {requestedVersion}"
+    else:
+        assert not matches, f"Expected {candidateVersion} to NOT match {requestedVersion}"
```
jmarrec, line 127: Test shows the way satisfiesVersion works

aqt/archives.py
```diff
@@ -36,18 +38,40 @@ def __init__(self, version, target, arch, os_name):
         self.arch = arch
         self.os_name = os_name
 
+    def __str__(self):
+        print(f'TargetConfig(version={self.version}, target={self.target}, '
+              f'arch={self.arch}, os_name={self.os_name}')
+
+    def __repr__(self):
+        print(f'({self.version}, {self.target}, {self.arch}, {self.os_name})')
+
```
jmarrec, line 47: Adding __repr__ and __str__ for easier debugging: you can just print objects (__str__) or lists of objects (__repr__ is used)

aqt/archives.py
```diff
@@ -81,7 +106,9 @@ def _get_archives(self):
         # Get packages index
         if self.version.major == 6 and self.target == "android":
             arch_ext = ["_armv7/", "_x86/", "_x86_64/", "_arm64_v8a/"]
-        elif self.version in SimpleSpec(">=5.13.0,<6.0") and self.target == "desktop":
+        elif ((self.target == 'desktop') and
+              (self.version >= Version('5.13.0')) and
+              (self.version < Version('6.0.0'))):
```
jmarrec, line 111: port from semantic_version to packaging.version, an example

aqt/archives.py
```diff
@@ -92,10 +119,10 @@ def _get_archives(self):
                 self.target,
                 self.version.major,
                 self.version.minor,
-                self.version.patch,
+                self.version.micro,
                 ext,
             )
-            update_xml_url = "{0}{1}Updates.xml".format(self.base, archive_path)
+            update_xml_url = posixpath.join(self.base, archive_path, "Updates.xml")
```
jmarrec, line 125: prefer posixpath to avoid any problems with `/` separators and co

aqt/archives.py
```diff
@@ -199,16 +227,20 @@ def _get_archives(self):
...
         target_packages.append(
             "qt.{0}{1}{2}.{3}".format(
-                self.version.major, self.version.minor, self.version.patch, self.arch
+                self.version.major, self.version.minor, self.version.micro, self.arch
             )
         )
         target_packages.extend(self.mod_list)
+        self.logger.debug('QtArchives::_get_archives')
+        self.logger.debug(f'{update_xml_url=}')
+        self.logger.debug(f'{target_packages=}')
```
jmarrec, line 242: perhaps shoudl remove, dunno

aqt/archives.py
```diff
@@ -36,18 +38,40 @@ def __init__(self, version, target, arch, os_name):
...
-    def __init__(self, name, archive_url, archive, package_desc, hashurl):
+    def __init__(self, name: str , archive_url: str, archive: str,
+                 package_desc: str, hashurl: str,
+                 version: Optional[Version] = None):
         self.name = name
         self.url = archive_url
         self.archive = archive
         self.desc = package_desc
         self.hashurl = hashurl
+        self.version = version
```
jmarrec, line 62: I store a version on the package. 
In the case of tools, there might be several matches. We want to filter on the max version to return a single

aqt/archives.py
```diff
@@ -380,40 +441,69 @@ def _parse_update_xml(self, archive_url, target_packages):
...
+            full_version = packageupdate.find("Version").text
+            # keep only '4.1.1' in '4.1.1-20210526113'
+            parsed_version = packaging.version.parse(full_version.split('-')[0])
+            versions.append((name, str(parsed_version)))
+
+            # Filter out on version first
+            parsed_version = packaging.version.parse(full_version.split('-')[0])
+
+            if self.version:
+                if not satifiesVersion(self.version, parsed_version):
```
jmarrec, line 458: This is the main change.

aqt/archives.py
```diff
@@ -380,40 +441,69 @@ def _parse_update_xml(self, archive_url, target_packages):
...
-                        QtPackage(name, package_url, archive, package_desc, hashurl)
-                    )
+
+            # Filter out on arch next
+            if self.arch and name != self.arch:
+                self.logger.warning(
+                    f'Arch of {name} is different from requested '
+                    f'arch {self.arch} -- skip.'
+                )
+                continue
```
jmarrec, line 472: Filter on arch next, if arch is given

aqt/archives.py
```diff
@@ -380,40 +441,79 @@ def _parse_update_xml(self, archive_url, target_packages):
...
+            self.logger.debug(f'ToolArchives: {self.archives}')
+        else:
+            msg = f"{self} did not match any available tools. "
+            if versions:
+                msg += 'Available tools\n'
+                msg += f'[(version, arch)] = {versions}'
+            else:
+                url = posixpath.join(archive_url, "Updates.xml")
+                msg += f"Zero tools where found to be available in {url}"
+            raise NoPackageFound(msg)
```
jmarrec, line 506: If none matches, raise with a helpful message

```
Arch of qt.tools.ifw.41 is different from requested arch qt.tools.ifw.40 -- skip.
Traceback (most recent call last):
 [...]
aqt.exceptions.NoPackageFound: ToolArchives(tool_name=tools_ifw, version=4.1, arch=qt.tools.ifw.40) did not match any available tools. Available tools
[(version, arch)] = [('qt.tools.ifw.41', '4.1.1')]

```

aqt/archives.py
```diff
@@ -380,40 +441,79 @@ def _parse_update_xml(self, archive_url, target_packages):
...
+
+        if len(self.archives) > 1:
+            self.logger.info(
+                f"Found {len(self.archives)} potential archives. "
+                "Returning the max.")
+            archive_max = max(self.archives,
+                              key=lambda package: package.version
+                              if package.version
+                              else packaging.version.Version('0'))
+            self.archives = [archive_max]
```
jmarrec, line 516: If more than one, only keep the max one.

jmarrec (commented):
added review comments as a walkthrough of my changes

aqt/installer.py
jmarrec: Hum this addition is this resulting in the Azure failures: https://dev.azure.com/miurahr/github/_build/results?buildId=4250&view=logs&j=e9afaa34-1a0f-534e-78fc-fae528ccd915&t=55eb6c4a-c635-53c7-061a-ddccec759601&l=24

```
python -m aqt src 6.1.0 linux desktop --archives qt
```

miurahr:
You are doing several different things at once.
Could you split PR?

jmarrec:
> You are doing several different things at once.
> Could you split PR?

Can you clarify what you would want separated please?

The majority of the changes I am doing in other areas than "tool" is because I changed `semantic_version` to `packaging.version`. Which was something I actually needed, since semantic_version will refuse a string like '4.0' and throw. 
850a369df29ceedc5426d386b648fdb81cd9675a was perhaps ill-advised, I commented it out for now, I can revert it completely depending on CI outcome

ddalcino:
> The majority of the changes I am doing in other areas than "tool" is because I changed `semantic_version` to `packaging.version`. Which was something I actually needed, since semantic_version will refuse a string like '4.0' and throw.

I think you might be better off using `semantic_version` instead of `packaging.version`. It already does everything you're trying to accomplish with `helper.satisfiesVersion`, and it also parses versions like "4.1.1-202105261130" without any problem.

If you want to accept strings like '4.0' as user input (for example, the user wants version '4.0' but doesn't care about the rest of the version number), you can use semantic_version.SimpleSpec. For example:

```python
Version('4.1.1-202105261130') in SimpleSpec('4.1')     # => True
Version('4.1.1-202105261130') in SimpleSpec('4.0')     # => False
Version('4.1.1-202105261130') in SimpleSpec('4')       # => True
```

jmarrec:
@ddalcino packaging is more likely to be installed already since it's used by setuptools though.

I didn't realize SimpleSpec could work because I also thought I would support versions like '4.x' originally and `SimpleSpec('4.x')` throws (later realized LegacyVersion doesn't support '.release' method).

I don't really care either way as long as partial versions are working.

ddalcino:
This project only recently switched from `packaging.version` to `semantic_version.Version` a couple of weeks ago; see #260 and #263. IIRC, we decided that it does a better job parsing version strings, and it's a more capable library in general. If you have a good reason as to why we should switch back, I would love to hear it. Maybe that could be a separate issue/PR?

> @ddalcino packaging is more likely to be installed already since it's used by setuptools though.

`semantic_version` is listed as a required dependency in `setup.cfg`, so it will be installed automatically when users install `aqt`. Yes, it is an additional dependency, and it weighs in at 50K disk space according to `du`. It's also popular enough that many users will already have it installed.

> I didn't realize SimpleSpec could work because I also thought I would support versions like '4.x' originally and SimpleSpec('4.x') throws (later realized LegacyVersion doesn't support '.release' method).

If you're trying to make `aqt tool` accept input like that, I really think you should at least take a look at SimpleSpec at https://pypi.org/project/semantic-version/. SimpleSpec supports versions like '4.*' and '>=4' by default:

```python
Version('4.1.1-202105261130') in SimpleSpec('4.*')    # => True
Version('4.1.1-202105261130') in SimpleSpec('>=4')    # => True
Version('4.1.1-202105261130') in SimpleSpec('>=4.2')  # => False
```

I suspect that attempting to re-create this kind of behavior using `packaging.version` and your own code will be outside the scope of this PR.

jmarrec:
Ok I see your point, especially given the (recent) history. I will revert those changes back to semantic_version when I get some free time then

jmarrec:
I will close this PR, and break it into several chunks

jmarrec:
First PR: https://github.com/miurahr/aqtinstall/pull/287

PR #292, opened 2021-07-01
Fix issues in getUrl

ddalcino:
This PR fixes several different issues in `helper.getUrl`. Please see the individual commit messages for details and the reasoning behind each change.

If you'd like, I can split this into four separate PRs, because this change fixes four separate problems. I just thought it would be easier to handle them all at once, since they're all in the same function.


miurahr:
Thank you!

PR #299, opened 2021-07-06
Allow tool sub-command to specify tool_name without version

miurahr:
Improve tool command; user can do such as follows;

- `aqt tool windows desktop tools_openssl_x64`
- `aqt tool windows android tools_qt3dstudio_runtime_240`
- `aqt tool windows desktop tools_conan`
- `aqt tool windows desktop tools_vcredist`
- `aqt tool linux desktop tools_ifw`
- `aqt tool mac desktop  tools_ninja`

Implement #282 and fix #233

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

PR #301, opened 2021-07-06
Validate version strings more consistently

ddalcino:
This removes the `exit(1)` call from the helper function `Cli._validate_version_str`, and makes that function return `True` or `False` instead. This puts the responsibilities of writing error messages and calling `exit` on the caller.

This change allows `Cli.run_list` to call `Cli._validate_version_str`, the same way that `run_install` and `run_src_doc_examples` do.

As a side effect, this allows `tests/test_cli.test_cli_invalid_version()` to pass, as discussed in my comment on #299.

This solution allows the test to pass, and I think it's a better solution than simply changing the test.

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -188,7 +188,7 @@ class ListCommand:
             return 0
         except CliInputError as e:
             self.logger.error("Command line input error: {}".format(e))
-            exit(1)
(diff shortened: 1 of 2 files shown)


miurahr:
Thank you!

PR #303, opened 2021-07-09
WIP: aqt tool improvements

ddalcino:
This allows commands such as `aqt tool windows desktop tools_openssl_x64 --spec ">=1.1"`.

This adds a command line argument to `aqt tool` called `--spec` that allows the user to specify constraints on the version number of the tool being installed, using a SimpleSpec. When `arch` is not specified, `aqt tool` will install the highest version of the tool available that fits the specified SimpleSpec. This argument is '*' by default, so if neither `arch` nor `--spec` is specified, `aqt tool` will select the highest version of the tools available in `tool_name`.

This feature will work on all tools, but it will not necessarily produce the intended results. Some `tool_names` include several different tools whose versions are not comparable. For example, `tools_conan` includes Conan 1.33 and Conan.cmake 0.16.0. These are completely different tools with different version histories. To install them properly, it only makes sense to install both variants `qt.tools.conan.cmake` and `qt.tools.conan` by name. `--spec` would only install one of them. Additionally, `tools_vcredist` and `tools_mingw` come in 32-bit and 64-bit flavors, and this tool does not distinguish between them.

So far, only two tests have been added for this feature, but I think they cover the most important parts of this code. However, this will require some additions to the Azure pipelines tests that have not been implemented yet.

Addresses #233 and #282.

### DIFF
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -297,6 +297,30 @@ def to_version(qt_ver: str) -> Version:
     )
 
 
+def to_version_permissive(version_str: str) -> Version:
+    """Converts a version string with dots (5.X.Y, etc) into a semantic version.
+    If the version omits either the patch or minor versions, they will be filled in with zeros,
+    and the remaining version string becomes part of the prerelease component.
+    If the version cannot be converted to a Version, a ValueError is raised.
+
+    This function is intended to be used on Version tags in an Updates.xml file.
+
+    to_version_permissive('1.33.1-202102101246') => Version('1.33.1-202102101246')
+    to_version_permissive('1.33-202102101246') => Version('1.33.0-202102101246')    # tools_conan
+    to_version_permissive('2020-05-19-1') => Version('2020.0.0-05-19-1')            # tools_vcredist
+    """
+    match = re.match(r"^(\d+)(\.(\d+)(\.(\d+))?)?(.*)$", version_str)
+    if not match:
+        raise ValueError("Invalid version detected: '{}'".format(version_str))
+
+    major, dot_minor, minor, dot_patch, patch, prerelease = match.groups()
+    return Version(
+        "{}.{}.{}{}".format(
+            major, minor if minor else 0, patch if patch else 0, prerelease
+        )
+    )
+
+
 def get_semantic_version(qt_ver: str, is_preview: bool) -> Optional[Version]:
     """Converts a Qt version string (596, 512, 5132, etc) into a semantic version.
     This makes a lot of assumptions based on established patterns:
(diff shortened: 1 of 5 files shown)


aqt/helper.py
```diff
@@ -297,6 +297,30 @@ def to_version(qt_ver: str) -> Version:
...
+    """Converts a version string with dots (5.X.Y, etc) into a semantic version.
+    If the version omits either the patch or minor versions, they will be filled in with zeros,
+    and the remaining version string becomes part of the prerelease component.
+    If the version cannot be converted to a Version, a ValueError is raised.
+
+    This function is intended to be used on Version tags in an Updates.xml file.
+
+    to_version_permissive('1.33.1-202102101246') => Version('1.33.1-202102101246')
+    to_version_permissive('1.33-202102101246') => Version('1.33.0-202102101246')    # tools_conan
+    to_version_permissive('2020-05-19-1') => Version('2020.0.0-05-19-1')            # tools_vcredist
```
ddalcino, line 310: This particular behavior might be a mistake. This is meant to handle version strings that are actually dates, as in qt.tools.windows_kits_debuggers and qt.tools.vcredist, as mentioned in my comment in #282. It might be better to use Date objects in this case, and then rely on other code to handle exceptions in the case that a Date is being compared with a Version. This will be an issue in the case of qt.tools.windows_kits_debuggers (which is in tools_qtcreator); most of the tools in that directory have Versions, but the debugger kit has a Date. IMHO in this case, and the tools_vcredist case, the user really should be using the 'tool variant name' parameter to install the tool, since the 'tool variant name' is stable and not incremented with each new version.

Maybe the `--spec` parameter should only be allowed for tools like `tools_ifw` that have 'tool variant name's that change constantly?

ddalcino:
I think that this PR might be a mistake. At present, the only tool for which the `--spec` argument has any relevance is QtIFW. I've taken a look at all the other tools, and none of them appears to have an auto-incrementing version number in the module name, other than QtIFW. If a user were to use `--spec` on any of the other tools, they most likely will not get what they wanted. I think that our users would be better served by being told to look up the modules in an Updates.xml file, or using the tool in #304, to figure out which modules they want to install.

In the case of QtIFW, this PR might be more useful. However, it looks like https://download.qt.io/online/qtsdkrepository only keeps one or two copies of QtIFW around at a time, and using the `--spec` argument will always fail eventually, unless you use '*'. I don't think there's much point in merging this.

The only really useful think that this PR does, for QtIFW, is choose the newest version available, rather than simply picking the first item in the list. For that reason, I'm not going to delete this branch.

PR #305, opened 2021-07-11
Introduce metadata package

miurahr:
Refactoring Versions and ListCommand class

- Introduce metadata module
  that has Version, Versions and ListCommand class

- Introduce custom Version class
    
    A Version class override semantic_version.Version class
    that accept qt versions and tools versions


miurahr:
@ddalcino I'd like to introduce new module `metadata` that has your improvements and your copyright.
Also I've done several refactoring on it.

- Introduce Version class inherited from semantic_version.Version class that accept Qt versions and other tools versions, that has custom parser
- Introduce `Versions.__format__()` method to help output of object values that renamed from `pretty_print()` 
- other improvements
 
Could you, @ddalcino , review for it?

tests/test_helper.py
```diff
@@ -102,11 +102,17 @@ def test_helper_downloadBinary_sha256(tmp_path, monkeypatch):
...
     [
-        ("1.33.1", Version("1.33.1")),
-        ("1.33.1-202102101246", Version("1.33.1-202102101246")),
-        ("1.33-202102101246", Version("1.33.0-202102101246")),
-        ("2020-05-19-1", Version("2020.0.0-05-19-1")),
+        ("1.33.1", Version(major=1, minor=33, patch=1)),
+        (
+            "1.33.1-202102101246",
+            Version(major=1, minor=33, patch=1, build=("202102101246",)),
+        ),
```
miurahr, line 109: Version.permissive() handles "-<datetime>" version string as a build version.

aqt/metadata.py
```diff
@@ -0,0 +1,657 @@
...
+        elif architectures_ver:
+            self.request_type = "architectures"
+            self._action = lambda: self.fetch_arches(
+                self._to_version(architectures_ver)
+            )
+        else:
+            self.request_type = "versions"
+            self._action = self.fetch_versions
+
+    def action(self) -> Union[List[str], Versions]:
```
ddalcino, line 347: This should probably have the type hint `-> Union[List[str], Versions, Table]`, because `Table` is what `action` returns in the case of `aqt list tools <host> <target> --tool-long <toolname>`. I forgot to add the Table type to the type hint for #304, and I really should have.

aqt/metadata.py
```diff
@@ -0,0 +1,657 @@
...
+                    "No {} available for this request.".format(self.request_type)
+                )
+                self.print_suggested_follow_up(self.logger.info)
+                return 1
+            if self.archive_id.is_tools():
+                print("\n".join(output))
+            elif isinstance(output, Versions):
+                print(f"{output}")
+            else:
+                print(" ".join(output))
```
ddalcino, line 364: I would be very careful about this kind of change. The Azure Pipeline is failing in this section (https://dev.azure.com/miurahr/github/_build/results?buildId=4344&view=logs&j=3b077b59-19b3-5c3c-d9ea-4f2d359714ca&t=fcf4a360-d878-547e-2092-53786c55870d&l=56), because output is of type `ListCommand.Table`, and `"\n".join(output)` doesn't work on that object type.

If you're going to use this approach to printing output, you will need to account for that case as well.

aqt/metadata.py
```diff
@@ -0,0 +1,657 @@
...
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
+from xml.etree import ElementTree as ElementTree
+
+import bs4
+from semantic_version import SimpleSpec as SementicSimpleSpec
```
ddalcino, line 42: I think you mean `SemanticSimpleSpec`?

aqt/metadata.py
```diff
@@ -0,0 +1,657 @@
...
+            msg = "Please use '{}' to check what tools are available.".format(base_cmd)
+            printer(msg)
+        elif self.filter_minor is not None:
+            msg = "Please use '{}' to check that versions of {} exist with the minor version '{}'".format(
+                base_cmd, self.archive_id.category, self.filter_minor
+            )
+            printer(msg)
+        elif self.request_type in ("architectures", "modules", "extensions"):
+            msg = "Please use '{}' to show versions of Qt available".format(base_cmd)
+            printer(msg)
```
ddalcino, line 657: 657 lines? You're right, it really did need its own file! Thank you for the refactor!

aqt/metadata.py
```diff
@@ -0,0 +1,657 @@
...
+
+from aqt.exceptions import ArchiveConnectionError, ArchiveDownloadError, CliInputError
+from aqt.helper import Settings, getUrl, xml_to_modules
+
+
+class SimpleSpec(SementicSimpleSpec):
+    pass
+
+
+class Version(SemanticVersion):
```
ddalcino, line 54: This override was a really good idea, thank you!

aqt/metadata.py
miurahr: Yes it is typo.

ddalcino (changes requested):
This looks really good. I think you just need to work out the issue with the ListCommand.Table failing to print, and then it's ready to merge as far as I'm concerned.

ddalcino (approved):
Looks great!

Not sure if you meant to change the behavior of `aqt list tools`, but fyi it now prints them as a space-separated list on a single line, rather than one tool per line.

aqt/metadata.py
miurahr: It use format when object supports, otherwise it use `" ".join(output)` which uses iterator of output.
It is better that classes supports iterator but Table does not support.

miurahr:
> Not sure if you meant to change the behavior of aqt list tools, but fyi it now prints them as a space-separated list on a single line, rather than one tool per line.

Thank you for it. I adjusted it now.

Issue #308, opened 2021-07-12
`aqt list tools ... --tool-long tools_qt3dstudio_runtime` gives unreadable output


ddalcino:
Here's the output for `python -m aqt list tools windows desktop --tool-long tools_qt3dstudio_runtime`:

```
                Tool Variant Name                          Version          Release Date                 Display Name                                                                                                                                                          Description                                                                                                                                         
===================================================================================================================================================================================================================================================================================================================================================================================================================================
qt.tools.qt3dstudioruntime.5110                     5.11.0-0-201806121305   2018-06-12     Qt 3D Studio Runtime 2.0.0 for Qt 5.11.0   Qt 3D Studio Runtime provides C++ and QML APIs for integrating Qt 3D Studio scenes into Qt applications.<br><br>This component is available under commercial licenses from The Qt Company, or under GPL v3. For open source use, please note the additional requirements compared to LGPL v3.
qt.tools.qt3dstudioruntime.5110.examples            5.11.0-0-201806121305   2018-06-12     Qt 3D Studio Runtime 5.11.0 examples       Qt 3D Studio Runtime 5.11.0 examples.                                                                                                                                                                                                                                                        
qt.tools.qt3dstudioruntime.5110.win32_msvc2015      5.11.0-0-201806121305   2018-06-12     Qt 3D Studio Runtime 5.11.0                Qt 3D Studio Runtime provides C++ and QML APIs for integrating Qt 3D Studio scenes into Qt applications.<br><br>This component is available under commercial licenses from The Qt Company, or under GPL v3. For open source use, please note the additional requirements compared to LGPL v3.
qt.tools.qt3dstudioruntime.5110.win64_msvc2015_64   5.11.0-0-201806121305   2018-06-12     Qt 3D Studio Runtime 5.11.0                Qt 3D Studio Runtime provides C++ and QML APIs for integrating Qt 3D Studio scenes into Qt applications.<br><br>This component is available under commercial licenses from The Qt Company, or under GPL v3. For open source use, please note the additional requirements compared to LGPL v3.
[...]

PR #309, opened 2021-07-13
Refactoring class structure: Factory class,  Data and consumer 

miurahr:
Enhancement for class design;

There are now three classes to hold metadata for `list` command

- `Versions` class: hold Qt versions information
- `ToolData` class: hold Tool names and versions
- `ArchiveId` class: common archive metadata

and factory class

- `ListCommand` class: produce the metadata of Qt download site

There is also a metadata for `install` command

- `List[QtPackage]`

and factory class

- `QtArchives` class

There is a public function;

- `show_list` : print listing of metadata for UI, a body of list sub-command
- `run_installer` : consume List[QtPackage] and install packages

private function;

- suggested_follow_up: show help

### DIFF
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -25,7 +25,17 @@ import posixpath
 import random
 import re
 from logging import getLogger
-from typing import Dict, Generator, Iterable, Iterator, List, Optional, Tuple, Union
+from typing import (
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 from xml.etree import ElementTree as ElementTree
 
 import bs4
@@ -616,14 +626,14 @@ class ListCommand:
         return "{} with minor version {}".format(self.archive_id, self.filter_minor)
 
 
-def suggested_follow_up(meta: ListCommand) -> str:
+def suggested_follow_up(meta: ListCommand, printer: Callable[[str], None]) -> None:
     """Makes an informed guess at what the user got wrong, in the event of an error."""
     base_cmd = "aqt {0.category} {0.host} {0.target}".format(meta.archive_id)
     if meta.archive_id.extension:
-        msg = "Please use '{} --extensions <QT_VERSION>' to list valid extensions.".format(
+        msg = "Please use '{} --extensions <QT_VERSION>' to list valid extensions.\n".format(
             base_cmd
         )
-        return msg
+        printer(msg)
 
     if meta.archive_id.is_tools() and meta.request_type == "tool variant names":
         msg = "Please use '{}' to check what tools are available.".format(base_cmd)
@@ -633,16 +643,16 @@ def suggested_follow_up(meta: ListCommand) -> str:
         )
     elif meta.request_type in ("architectures", "modules", "extensions"):
         msg = "Please use '{}' to show versions of Qt available".format(base_cmd)
-    return msg
+    printer(msg)
 
 
 def show_list(meta: ListCommand) -> int:
-    logger = getLogger()
+    logger = getLogger("aqt.list")
     try:
         output = meta.action()
         if not output:
             logger.info("No {} available for this request.".format(meta.request_type))
-            logger.info(suggested_follow_up(meta))
+            suggested_follow_up(meta, logger.info)
             return 1
         if isinstance(output, Versions):
             print(format(output))
@@ -658,5 +668,5 @@ def show_list(meta: ListCommand) -> int:
         return 1
     except (ArchiveConnectionError, ArchiveDownloadError) as e:
         logger.error("{}".format(e))
-        logger.error(suggested_follow_up(meta))
+        suggested_follow_up(meta, logger.error)
         return 1
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -371,7 +371,7 @@ class MetadataFactory:
             self.request_type = "versions"
             self._action = self.fetch_versions
 
-    def action(self) -> Union[List[str], Versions, ToolData]:
+    def getList(self) -> Union[List[str], Versions, ToolData]:
         return self._action()
 
     def fetch_modules(self, version: Version) -> List[str]:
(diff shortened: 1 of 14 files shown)


aqt/metadata.py
```diff
@@ -641,23 +615,48 @@ def describe_filters(self) -> str:
...
-            printer(msg)
+
+def suggested_follow_up(meta: ListCommand) -> str:
+    """Makes an informed guess at what the user got wrong, in the event of an error."""
+    base_cmd = "aqt {0.category} {0.host} {0.target}".format(meta.archive_id)
+    if meta.archive_id.extension:
+        msg = "Please use '{} --extensions <QT_VERSION>' to list valid extensions.".format(
+            base_cmd
+        )
+        return msg
```
ddalcino, line 626: You've made a control flow change here, and I'm not sure you meant to.

The previous implementation was capable of printing multiple suggestions for the same command: one suggestion to print available extensions, and one suggestion that depends on the request type.

For example: a user types `aqt list qt6 windows android --extension x86 --modules 6.2.2`. This command will fail, because https://download.qt.io/online/qtsdkrepository/windows_x86/android/qt6_622_x86/Updates.xml does not exist.  Unless we read the html page at https://download.qt.io/online/qtsdkrepository/windows_x86/android/, we cannot know if the user got the extension wrong or the version wrong. My solution was to print both suggestions: `aqt list --extensions`, and `aqt list` to print available versions. This change would only print the `aqt list --extensions` message, and that happens to be the wrong suggestion in this case.

It now occurs to me that the right solution to this problem might be to go read the html page and figure out which part(s) the user got wrong, and then print to the console "The version 6.2.2 doesn't exist; these are the versions available: 6.0.0 ...", instead of "Try this command to list versions available".

aqt/metadata.py
```diff
@@ -641,23 +615,48 @@ def describe_filters(self) -> str:
...
-        elif self.filter_minor is not None:
-            msg = "Please use '{}' to check that versions of {} exist with the minor version '{}'".format(
-                base_cmd, self.archive_id.category, self.filter_minor
-            )
-            printer(msg)
-        elif self.request_type in ("architectures", "modules", "extensions"):
-            msg = "Please use '{}' to show versions of Qt available".format(base_cmd)
-            printer(msg)
+
+def suggested_follow_up(meta: ListCommand) -> str:
```
ddalcino, line 619: This is probably a good idea!

aqt/metadata.py
```diff
@@ -270,24 +260,45 @@ def __str__(self) -> str:
...
 
 
-class Table:
-    def __init__(self, head: List[str], rows: List[List[str]], max_width: int = 0):
-        # max_width is set to 0 by default: this disables wrapping of text table cells
-        self.head = head
-        self.rows = rows
-        self.max_width = max_width
+class ToolData:
+    """A data class hold tool details."""
```
ddalcino, line 264: I like the idea of this refactor a lot.

aqt/metadata.py
ddalcino: Now I wish I had added tests for this function!

aqt/metadata.py
```diff
@@ -270,24 +260,45 @@ def __str__(self) -> str:
...
+    @property
+    def rows(self):
+        keys = ("Version", "ReleaseDate", "DisplayName", "Description")
+        return [
+            [name, *[content[key] for key in keys]]
+            for name, content in self.tool_data.items()
+        ]
 
 
 class ListCommand:
```
miurahr, line 304: It may be better to change the name to express what it is doing.

Now `ListCommand` class has only a factory functions to retrieve Update.xml and produce metadata(`List[str]`, `Versions` or `ToolData`).
Here is `ListCommand` as a factory class, `ToolData` is a data class, and `show_list` function is a consumer.

There is `QtArchives` as a factory class, `List[QtPackages]` is a data, and `call_installer` function is a consumer.

There are more spaces to improve a structure of classes.

aqt/metadata.py
miurahr: I don't intend to change a behavior. It would be some mistakes here.

aqt/metadata.py
```diff
@@ -641,23 +625,50 @@ def describe_filters(self) -> str:
...
+    try:
+        output = meta.action()
+        if not output:
+            logger.info("No {} available for this request.".format(meta.request_type))
+            suggested_follow_up(meta, logger.info)
+            return 1
+        if isinstance(output, Versions):
+            print(format(output))
+        elif isinstance(output, ToolData):
+            print(format(output, "{:t}"))  # can set width "{:100t}"
```
miurahr, line 662: This mechanism may help solving the issue #308
You can add more format spec to control output, if wants.

aqt/metadata.py
```diff
@@ -270,24 +270,45 @@ def __str__(self) -> str:
...
-            table.add_rows(self.rows, header=False)
-            return table.draw()
-        elif format_spec == "s":
+        if format_spec == "{s}":
             return str(self)
+        if format_spec == "":
+            max_width: int = 0
         else:
-            raise ValueError()
+            match = re.match(r"\{(.*):(\d*)t\}", format_spec)
```
miurahr, line 293: Accept `{:t}`,  and width spec such as `{:0t}` or `{:100t}`

aqt/metadata.py
ddalcino: I agree. Maybe we could rename `ListCommand` to something like `MetadataFactory`? Maybe `ListCommand.action()` could be renamed `MetadataFactory.getMetadata()`?

aqt/metadata.py
ddalcino: I don't think you have to throw out your changes entirely. I like the idea of `suggested_follow_up` returning a string that the caller could print; it just needs to be able to append multiple suggestions to the return value.

miurahr: I've reverted the function.

aqt/metadata.py
miurahr: I'd like to just move from a class member to an individual public function.

aqt/metadata.py
miurahr: done

miurahr:
@ddalcino thank you for detailed review.
I've squashed and merged it.

PR #314, opened 2021-07-14
Extension validation in ListCommand

ddalcino:
In the current implementation, if a user runs `aqt list qt6 windows android --modules 6.2.0`, without an extension, aqt prints this:
```
No modules available for this request.
Please use 'aqt qt6 windows android' to show versions of Qt available
```

This happens because all the Qt6 Android modules require an extension for the processor architecture. The current error message does nothing to indicate what the problem is. The user's request seems perfectly reasonable

# What this PR does:
This change adds a function that checks for obvious misuses and omissions of the `--extension` flag, according to these rules:

1. On Qt6 for Android, an extension for Android processor architecture ("x86_64", "x86", "armv7", "arm64_v8a") is required.
2. On any platform other than Android, or on Qt5, the extensions for Android processor architecture are not allowed.
3. The "wasm" extension only works on desktop targets for Qt 5.13-5.15.

This validation function is only called on requests for 'modules' and 'architectures', because these are the only places where the version of Qt is known and the extension matters. Hopefully, this change will reduce confusion among users.

### DIFF
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -214,6 +214,7 @@ class ArchiveId:
         "armv7",
         "arm64_v8a",
     )
+    EXTENSIONS_REQUIRED_ANDROID_QT6 = "x86_64", "x86", "armv7", "arm64_v8a"
 
     def __init__(self, category: str, host: str, target: str, extension: str = ""):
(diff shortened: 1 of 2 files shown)


miurahr:
LGTM

PR #318, opened 2021-07-15
Add GH Action to autogenerate `combinations.json`

### LINKED ISSUE #307: `aqt/combinations.json` can be generated using `aqt list`
I've noticed that over time, `aqt/combinations.json` loses sync with the Qt repository, and needs to be updated manually. I think that it should be possible, using `aqt list`, to re-generate/update the whole file using a program. I think that I could do this, but I have some questions before I start:

1. Is this something that we want to add to the project?
2. Would it be a separate program, or would we add it as a subcommand to aqt, like `python -m aqt update-combos`?
3. Should it run in CI, and cause the build to fail if the current `aqt/combinations.json` file is not up to date?

Keep in mind that this command would necessarily make a whole lot of GET requests to the Qt repo:
```
3 hosts * 3 targets * (1 html page + N Updates.xml files)
```

I think it would make sense for the aqt developers to run this command periodically, but it might be a big problem if everyone who uses aqt decides to run it in their CI. The Qt people won't be happy if we use up an excessive amount of their bandwidth.

ddalcino:
This adds a script called `generate_combinations.py` that:
1. Runs ListCommand to generate a dictionary that approximates the contents of `combinations.json`. This step takes ~2 minutes to run on my machine, so I used `tqdm` to give some indication of progress.
2. Pretty-prints the dictionary to the console, in a way that mimics the formatting of `combinations.json`
3. Prints a detailed comparison of the program output with the actual `combinations.json`.

~~This script makes no attempt to check `new_archive`, since that is something that ListCommand cannot do.~~
~~**Edit:** I added a function to metadata.py that reads 'new_archive', so this tool can now reproduce the whole `combinations.json` file.~~
**Edit:** The 'new_archive' data is now inherited from the previous `combinations.json` file.

Fixes #307

### DIFF
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -47,4 +47,4 @@ jobs:
         branch: "update-combinations"
         path: "aqt/combinations.json"
         commit-message: "Update aqt/combinations.json"
-        author: "Qt Repo Watchbot <ddalcino@gmail.com>"
+        author: "Qt Repo Watchbot <qtrepowatchbot[bot]@users.noreply.github.com>"
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -41,8 +41,7 @@ jobs:
       with:
         title: "Update `aqt/combinations.json`"
         body: |
-          SUMMARY
-          The `aqt/generate_combinations.py` script has detected changes to the repo at https://download.qt.io.
+          The `ci/generate_combinations.py` script has detected changes to the repo at https://download.qt.io.
           This PR will update `aqt/combinations.json` to account for those changes.
 
           Posted from [the `check_combinations` action](https://github.com/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }})
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -5,7 +5,10 @@ on:
   # https://crontab.guru/once-a-month
   - cron: "0 0 1 * *"
 
+  workflow_dispatch:
   push:
+    paths:
+      - 'ci/**'
 
 jobs:
   check_combinations:
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -1,0 +1,47 @@
+name: "Check combinations.json"
+on:
+  schedule:
+  # Run at midnight on the first of every month
+  # https://crontab.guru/once-a-month
+  - cron: "0 0 1 * *"
+
+  push:
+
+jobs:
+  check_combinations:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: 20
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Build and install
+      run: |
+        python setup.py --version
+        python -m pip install ./ --user
+
+    - name: Check combinations.json
+      run: PYTHONPATH=$(pwd) python3 ci/generate_combinations.py --write --no-tqdm
+
+    - name: Commit and make pull request
+      uses: gr2m/create-or-update-pull-request-action@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        title: "Update `aqt/combinations.json`"
+        body: |
+          SUMMARY
+          The `aqt/generate_combinations.py` script has detected changes to the repo at https://download.qt.io.
+          This PR will update `aqt/combinations.json` to account for those changes.
+
+          Posted from [the `check_combinations` action](https://github.com/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }})
+
+        branch: "update-combinations"
+        path: "aqt/combinations.json"
+        commit-message: "Update aqt/combinations.json"
+        author: "Qt Repo Watchbot <ddalcino@gmail.com>"
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -28,6 +28,9 @@ jobs:
         python setup.py --version
         python -m pip install ./ --user
 
+    - name: Install generate_combinations dependencies
+      run: pip install -U jsoncomparison
+
     - name: Check combinations.json
       run: PYTHONPATH=$(pwd) python3 ci/generate_combinations.py --write --no-tqdm
 
--- a/ci/generate_combinations.py
+++ b/ci/generate_combinations.py
@@ -5,16 +5,7 @@ import json
 import logging
 import re
 from pathlib import Path
-from typing import (
-    Dict,
-    Generator,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    Union,
-)
+from typing import Dict, Generator, Iterator, List, Optional, Set, Tuple, Union
 
 from aqt.exceptions import ArchiveConnectionError, ArchiveDownloadError
 from aqt.helper import Settings, setup_logging
@@ -375,6 +366,7 @@ def get_tqdm(disable: bool):
         return lambda x: x
 
     from tqdm import tqdm as base_tqdm
+
     return lambda *a: base_tqdm(*a, disable=disable)
 
 
--- a/ci/generate_combinations.py
+++ b/ci/generate_combinations.py
@@ -3,9 +3,10 @@
 import argparse
 import json
 import logging
-import re
 from pathlib import Path
-from typing import Dict, Generator, Iterator, List, Optional, Set, Tuple, Union
+from typing import Dict, Generator, Iterator, List, Optional, Tuple, Union
+
+from jsoncomparison import NO_DIFF, Compare
 
 from aqt.exceptions import ArchiveConnectionError, ArchiveDownloadError
 from aqt.helper import Settings, setup_logging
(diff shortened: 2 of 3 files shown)


ddalcino:
Output for the most recent run: https://gist.github.com/ddalcino/5bf8b705f85b32bf026cfc2cf02640ff

aqt/generate_combinations.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+def is_blacklisted_tool(tool_name: str) -> bool:
+    for prefix in ("tools_qt3dstudio_",):
+        if tool_name.startswith(prefix):
+            return True
+    for suffix in ("_preview", "_early_access"):
+        if tool_name.endswith(suffix):
+            return True
+    return False
```
ddalcino, line 33: This is a strange thing to do: it skips a large number of tools. The tools that this skips are not tracked by `aqt/combinations.json`, and they constitute the majority of the HTTP requests that this tool makes if they are not skipped. In the future, maybe this behavior should be turned off.

ddalcino:
Pretty much anywhere I used `tqdm` in this script would be a good place to think about using concurrency. This script takes ~2 minutes to run on my machine, and most of that time is spend waiting on HTTP GET requests. Currently, all those requests are sent out in series; if all the requests were sent at once, I would expect the script to finish a whole lot quicker. Since this script will not be run by end users, I don't think that this idea is very urgent, so I will leave it for another PR.

ddalcino:
We should probably audit the output of this script to decide how it should be changed.

```
Comparing qt:
----------------------------------------
combinations.json['qt'] is missing these entries:
{'os_name': 'mac', 'target': 'android', 'arch': 'android_arm64_v8a'}
{'os_name': 'mac', 'target': 'android', 'arch': 'android_armv7'}
{'os_name': 'mac', 'target': 'android', 'arch': 'android_x86'}
{'os_name': 'mac', 'target': 'android', 'arch': 'android_x86_64'}
{'os_name': 'windows', 'target': 'desktop', 'arch': 'win64_msvc2013_64'}
{'os_name': 'windows', 'target': 'desktop', 'arch': 'win64_msvc2019_arm64'}
{'os_name': 'windows', 'target': 'winrt', 'arch': 'win64_msvc2015_winrt_x86'}
```
In the 'qt' section, the script picks out a couple of architectures that aren't present in `combinations.json`. I think that these architectures should be added to `combinations.json`, because they exist in the repo.

The script does not perform an exhaustive search for architectures, so it could be possible that there are a few that it's missing. The script runs the equivalent `aqt list qt5/6 <host> <target> --arch <version>` for all appropriate hosts and targets, and the following versions: Qt6 latest, Qt5 latest, 5.13.2, and 5.9.9. I chose versions 5.13.2 and 5.9.9 because when I left either of them out, the script omits a few architectures that are present in `combinations.json`. Each version of Qt supports a slightly different set of architectures on Windows. I doubt that it will be useful to run the script on all versions of Qt, but I think that's what it would take to make sure that the script doesn't omit any architectures.

```
Comparing tools:
----------------------------------------
program_output['tools'] is missing these entries:
{'os_name': 'linux', 'target': 'desktop', 'tool_name': 'tools_ifw', 'arch': 'qt.tools.ifw.40'}
{'os_name': 'linux', 'target': 'desktop', 'tool_name': 'tools_maintainance', 'arch': 'qt.tools.maintainance'}
{'os_name': 'linux', 'target': 'desktop', 'tool_name': 'tools_openssl_src', 'arch': 'qt.tools.openssl'}
{'os_name': 'mac', 'target': 'desktop', 'tool_name': 'tools_ifw', 'arch': 'qt.tools.ifw.40'}
{'os_name': 'mac', 'target': 'desktop', 'tool_name': 'tools_maintainance', 'arch': 'qt.tools.maintainance'}
{'os_name': 'windows', 'target': 'desktop', 'tool_name': 'tools_ifw', 'arch': 'qt.tools.ifw.40'}
{'os_name': 'windows', 'target': 'desktop', 'tool_name': 'tools_maintainance', 'arch': 'qt.tools.maintainance'}
```
These are tools that are present in `combinations.json` but this script cannot locate. The `tools_maintainance` items are not present because `maintenance` is misspelled in `combinations.json`. QtIFW 4.0 has been replaced by 4.1. The 'arch' for openssl_src was also updated.

This script also detects a huge list of tools that are not present in `combinations.json`. I added a blacklist to prevent the script from checking all of them; this cuts down on the time it takes to run the script drastically.

```
Comparing modules:
----------------------------------------
combinations.json['modules'] for Qt 5.12 is missing {'debug_info'}
program_output['modules'] for Qt 5.13 is missing {'qtquicktimeline'}
combinations.json['modules'] for Qt 5.13 is missing {'debug_info'}
combinations.json['modules'] for Qt 5.14 is missing {'qtwaylandcompositor', 'debug_info'}
combinations.json['modules'] for Qt 5.15 is missing {'qtwaylandcompositor', 'debug_info'}
combinations.json['modules'] for Qt 6.0 is missing {'debug_info'}
combinations.json['modules'] for Qt 6.1 is missing {'addons.qtdatavis3d', 'addons.qtscxml', 'addons.qtnetworkauth', 
  'addons.qtimageformats', 'debug_info', 'addons.qtcharts', 'addons.qtlottie', 'addons.qt3d', 'addons.qtvirtualkeyboard'}
combinations.json['modules'] for Qt 6.2 is missing {'addons.qtmultimedia', 'addons.qtscxml', 'addons.qtnetworkauth', 
  'addons.qtimageformats', 'addons.qtwebchannel', 'addons.qtcharts', 'addons.qtwebview', 'addons.qtpositioning', 
  'addons.qtserialport', 'addons.qtserialbus', 'addons.qtconnectivity', 'addons.qtwebengine', 'addons.qtlottie', 
  'addons.qtdatavis3d', 'addons.qtwebsockets', 'addons.qtremoteobjects', 'debug_info', 'addons.qtsensors',
  'addons.qt3d', 'addons.qtvirtualkeyboard'}
```
This part of the script just checks modules for Linux desktop, for the most recent release of each minor version. I thought it would be wasteful to check all three platforms here, but there are minor differences. Many releases of Qt have a 'debug_info' module on Linux and Windows, but not Mac. There may be other differences that I am not aware of.

I'm not sure why `combinations.json` includes 'qtquicktimeline' for Qt 5.13; I can't find any evidence of this module for 5.13.

I don't know what's going on with `combinations.json` omitting the modules that include the word 'addons'. We can filter them out if they are not appropriate to include in `combinations.json`. If `aqt` cannot install them, MetadataFactory should filter them out too.

```
Comparing versions:
----------------------------------------
program_output['versions'] is missing these entries:
['5.9']
combinations.json['versions'] is missing these entries:
['5.9.0', '6.1.2']
```
I think that Qt 6.1.2 was released more recently than the last time `combinations.json` was updated. The Qt '5.9.0' thing is related to #312. I think that as long as `aqt install 5.9.0` works okay, it makes more sense to use the whole version number.

```
Comparing new_archive:
----------------------------------------
program_output['new_archive'] is missing these entries:
['5.1.0', '5.2', '5.3.0', '5.3.1', '5.3.2', '5.4.0', '5.4.1', '5.4.2', '5.5.0', '5.5.1', '5.6.0', '5.6.1', '5.6.2', '5.6.3', '5.7.0', '5.7.1', '5.8.0']
```
This script does not attempt to locate 'new_archive' versions because I'm not sure what 'new_archive' is.

ddalcino:
```
Comparing new_archive:
----------------------------------------
program_output['new_archive'] is missing these entries:
['5.2']
combinations.json['new_archive'] is missing these entries:
['5.2.0', '5.2.1']
```
I cannot explain the discrepancy here. There's clearly a version of Qt at 'new_archive' for versions 5.2.0 and 5.2.1: 
* https://download.qt.io/new_archive/qt/5.2/5.2.0/
* https://download.qt.io/new_archive/qt/5.2/5.2.1/

aqt/generate_combinations.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+def alphabetize_modules(combos: Dict):
+    for i, item in enumerate(combos["modules"]):
+        combos["modules"][i]["modules"] = sorted(item["modules"])
+
+
+if __name__ == "__main__":
+    Settings.load_settings()
+    logger = logging.getLogger("aqt.generate_combos")
```
miurahr, line 320: When introduce a new logger, please add it in logging.ini definition.
Current logger structure is

```
root -
       - aqt
                - main
                - archives
                - helper
                - installer
                - list
                - updater
```

ddalcino: Oops, good call!

@miurahr, I have a question about the loggers used in `aqt/metadata.py`. The `metadata.show_list` function uses the `aqt.list` logger, but `metadata.MetadataFactory` still uses the `aqt.archives` logger. All of the other loggers appear to have the same name as the file in which they are used. Would it be appropriate to rename the `aqt.list` logger to `aqt.metadata`, and use that logger in both `metadata.MetadataFactory` and `metadata.show_list`?

aqt/metadata.py
```diff
@@ -341,6 +341,26 @@ def _short_rows(self):
         ]
 
 
+def fetch_new_archive_versions(
```
miurahr, line 344: IMHO, it should not be implemented.

aqt/generate_combinations.py
```diff
@@ -0,0 +1,391 @@
...
+        _hash = record["os_name"], record["target"], record["arch"]
+        if _hash not in hashes:
+            all_records.append(record)
+            hashes.add(_hash)
+    for sorting_key in ("arch", "target", "os_name"):
+        all_records = sorted(all_records, key=lambda d: d[sorting_key])
+    return all_records
+
+
+def list_new_archive_versions() -> List[str]:
```
miurahr, line 163: new_archive (= old versions, offline installer , not supported) should not be into supported combo.

miurahr (changes requested):
Could you consider followings?

- Please don't include `new_archive` things
- Please put script in `ci` folder
- Only function that is common for aqt commands are in `aqt` package.

aqt/generate_combinations.py
```diff
@@ -0,0 +1,391 @@
...
+
+def open_pull_request():
+    """
+    $ git add aqt/combinations.json
+    $ git commit -m "Update aqt/combinations.json"
+    """
+    raise NotImplementedError()
+
+
+def main(is_make_pull_request: bool) -> int:
```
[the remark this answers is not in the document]
ddalcino: Yes, I can move it. I wasn't sure where to put it.

tests/test_list.py
```diff
@@ -368,3 +375,20 @@ def _mock(_, rest_of_url: str) -> str:
     sys.stdout.write(out)
     sys.stderr.write(err)
     assert expected_msg in err
+
+
+def test_list_new_archive(monkeypatch):
```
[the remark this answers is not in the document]
ddalcino: Ok. I was hoping to autogenerate the whole `combinations.json` file, but I can just read it from the existing file instead.

aqt/generate_combinations.py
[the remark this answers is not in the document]
ddalcino: Moved

tests/test_list.py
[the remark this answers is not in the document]
ddalcino: Removed

aqt/generate_combinations.py
ddalcino: removed

aqt/generate_combinations.py
miurahr: > I have a question about the loggers used in aqt/metadata.py. The metadata.show_list function uses the aqt.list logger, but metadata.MetadataFactory still uses the aqt.archives logger.

This inconsistency is just a bug, sorry.

.github/workflows/check-combinations.yml
```diff
@@ -0,0 +1,47 @@
+name: "Check combinations.json"
+on:
+  schedule:
+  # Run at midnight on the first of every month
+  # https://crontab.guru/once-a-month
+  - cron: "0 0 1 * *"
+
+  push:
```
ddalcino, line 8: This action runs once a month and on every push. When this is ready to merge, "push" should be removed: we don't want it running that often. Push is turned on for now for testing purposes only.

Also, we should revisit how often this should run. I think that once a month may be too infrequent; what do you think about once a week?

miurahr, line 8: I'd like to use  `workflow_dispatch:` instead of `push` 
It allow admin to run when it needed. It is convenient for test purpose.

.github/workflows/check-combinations.yml
```diff
@@ -0,0 +1,47 @@
...
+          SUMMARY
+          The `aqt/generate_combinations.py` script has detected changes to the repo at https://download.qt.io.
+          This PR will update `aqt/combinations.json` to account for those changes.
+
+          Posted from [the `check_combinations` action](https://github.com/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }})
+
+        branch: "update-combinations"
+        path: "aqt/combinations.json"
+        commit-message: "Update aqt/combinations.json"
+        author: "Qt Repo Watchbot <ddalcino@gmail.com>"
```
ddalcino, line 47: I wasn't sure what to put here for author, and I'm not sure I really want my email address in this source code. I had to put something in, otherwise the action fails.

.github/workflows/check-combinations.yml
miurahr: It is also worth considering
```
on:
  push:
     paths:
       - 'ci/**'
```

ci/generate_combinations.py
```diff
@@ -0,0 +1,398 @@
...
+    filename: Path,
+    is_use_pretty_print: bool = True,
+):
+    logger.info(f"Write file {filename}")
+    json_text = (
+        pretty_print_combos(combos)
+        if is_use_pretty_print
+        else json.dumps([combos], sort_keys=True, indent=2)
+    )
+    if filename.write_text(json_text, encoding="utf_8") == 0:
```
ddalcino, line 331: By default, this script writes the json file using my `pretty_print_combos` function, which is not well tested. It's possible that this code could write the wrong thing in the json file.

The `json.dumps` function is well tested, and I think it would work do a better job at printing the json file. However, with the parameter `indent=2`, it writes 870 lines of json data, which is not so easy to read. If you turn off `indent`, it just writes the whole object on a single line.

.github/workflows/check-combinations.yml
ddalcino: Those are really good ideas! I will do that!

ddalcino:
This is the most recent run of the `check-combinations.yml` action: https://github.com/ddalcino/aqtinstall/runs/3110395099?check_suite_focus=true

Here's the pull request opened by that action: https://github.com/ddalcino/aqtinstall/pull/3

Please note that the PR shows 3 commits: the first two are the commits that I made to make the `check-combinations` action, and the last was auto-generated. If the ddalcino:poc-generate-combinations branch had been merged into master before the action ran, the auto-generated PR would only show one commit, and one file changed: `aqt/combinations.json`.

.github/workflows/check-combinations.yml
miurahr: You can put dummy mail address here as like a dependabot.
``` dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com> ```

ci/generate_combinations.py
miurahr: When the action involved, no one doesn't read json directly. Otherwise, we will use the utility.
It is ok to use `json.dump` with indent.  Github help us to see difference in PR diff.

miurahr (approved):

ci/generate_combinations.py
miurahr: https://pypi.org/project/jsoncomparison/ may be usable to show diff.

.github/workflows/check-combinations.yml
ddalcino: Ok, I will try that!

ci/generate_combinations.py
ddalcino: > It is ok to use `json.dump` with indent. Github help us to see difference in PR diff.

Sounds good to me.

> https://pypi.org/project/jsoncomparison/ may be usable to show diff.

I tried using `jsoncomparison` to show the diff for this program's output, but I am not happy with the result. Here's a link to some recent output: https://gist.github.com/ddalcino/687515d367d543372fe5924e6f0c5372

I think `jsoncomparison` would be perfectly fine for checking whether a difference exists or not, and thereby deciding whether to open a PR. I think that the output shown in gist above will be useless to us, so I just won't print it. Github will be a lot better at showing us the diff anyway.

PR #323, opened 2021-07-16
Add `aqt.metadata.Versions.flattened()`

ddalcino:
This adds a function to Versions that returns a flattened list of Version objects. This is a spinoff from #318 that I needed for that feature, but I think it's outside the scope of that PR.

This also adds test coverage for the whole Versions class.


miurahr:
LGTM

PR #325, opened 2021-07-18
Refactor `suggested_follow_up`

ddalcino:
This PR splits `aqt.metadata.suggested_follow_up` into:
1. a function that returns a list of strings, and
2. a function that formats and prints those strings

This PR also adds test code for both functions.

Here is an example of what the formatter prints:
```
* Please use 'aqt list tools mac desktop --extensions <QT_VERSION>' to list valid extensions.
* Please use 'aqt list tools mac desktop' to check what tools are available.
```

### DIFF
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -20,7 +20,6 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import itertools
-import logging
 import operator
 import posixpath
 import random
@@ -28,7 +27,6 @@ import re
 import shutil
 from logging import getLogger
 from typing import (
-    Callable,
     Dict,
     Generator,
     Iterable,
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -20,6 +20,7 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import itertools
+import logging
 import operator
 import posixpath
 import random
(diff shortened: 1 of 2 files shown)


aqt/metadata.py
```diff
@@ -695,26 +695,33 @@ def describe_filters(self) -> str:
         return "{} with minor version {}".format(self.archive_id, self.filter_minor)
 
 
-def suggested_follow_up(meta: MetadataFactory, printer: Callable[[str], None]) -> None:
+def suggested_follow_up(meta: MetadataFactory) -> List[str]:
```
miurahr, line 698: It can be `print_suggested_follow_up()` and accept log level such as `logging.INFO` and `logging.ERROR`.
You can use logger as follows;
```
logger = logging.getLogger()
logger.log(logging.INFO, msg)
```
https://docs.python.org/ja/3/library/logging.html#logging.Logger.log

This allow you not to pass Callable.

ddalcino: How about calling it `log_suggested_follow_up(meta, log_level)`, since making that change would restrict it to outputting to the logger?

ddalcino: After making this change, the test `test_log_suggested_follow_up` passes when it is run on its own, but it fails when run with the rest of the tests. The error message says it failed because the logger file was closed while it was being written to: https://github.com/miurahr/aqtinstall/runs/3099834773?check_suite_focus=true#step:6:273.  

I think this is related to https://github.com/pytest-dev/pytest/issues/5502. That issue mentions some workarounds, but I haven't been able to get them to work.

miurahr: OK, I'd like to merge it with current version.

Issue #328, opened 2021-07-20
'aqt list tools' is broken - ValueError: Invalid version string: 'tools'


danryu:
Using aqt 1.2.4 on Linux.
Following documentation at https://aqtinstall.readthedocs.io/en/latest/cli.html 

```
test@hp14:~$ aqt list tools windows desktop 
aqtinstall(aqt) v1.2.4 on Python 3.9.5 [CPython GCC 10.3.0]
Traceback (most recent call last):
  File "/usr/local/bin/aqt", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/aqt/__init__.py", line 31, in main
    return cli.run()
  File "/usr/local/lib/python3.9/dist-packages/aqt/installer.py", line 675, in run
    result = args.func(args)
  File "/usr/local/lib/python3.9/dist-packages/aqt/installer.py", line 472, in run_list
    pl = PackagesList(qt_version, host, target, Settings.baseurl)
  File "/usr/local/lib/python3.9/dist-packages/aqt/archives.py", line 72, in __init__
    self.version = Version(version)
  File "/home/test/.local/lib/python3.9/site-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/home/test/.local/lib/python3.9/site-packages/semantic_version/base.py", line 311, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: 'tools'
test@hp14:~$ aqt list tools linux desktop 
aqtinstall(aqt) v1.2.4 on Python 3.9.5 [CPython GCC 10.3.0]
Traceback (most recent call last):
  File "/usr/local/bin/aqt", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/aqt/__init__.py", line 31, in main
    return cli.run()
  File "/usr/local/lib/python3.9/dist-packages/aqt/installer.py", line 675, in run
    result = args.func(args)
  File "/usr/local/lib/python3.9/dist-packages/aqt/installer.py", line 472, in run_list
    pl = PackagesList(qt_version, host, target, Settings.baseurl)
  File "/usr/local/lib/python3.9/dist-packages/aqt/archives.py", line 72, in __init__
    self.version = Version(version)
  File "/home/test/.local/lib/python3.9/site-packages/semantic_version/base.py", line 105, in __init__
[...]

PR #329, opened 2021-07-21
Link documentation for stable

miurahr:
work around #328

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@ Another Qt installer(aqt)
 =========================
 
 - Release: |pypi|
-- Documents: |docs|
+- Documentation: |docs|
 - Test status: |gha| and Coverage: |coveralls|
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -70,6 +70,15 @@ Requirements
 
 - Dependent libraries: requests, py7zr
 
+Documents
+---------
+
+There are versions of document
```
ddalcino, line 76: This line doesn't quite look right to me. I would prefer it said this:

```diff
- Documents
+ Documentation
  -------------------

- There are versions of document
+ There are two versions of the documentation: 
```

ddalcino: Also, above on line 5, we have the word "Documents" labelling a link to the documentation. Can we change that to the word "Documentation" as well?

miurahr: Thank you!

Actually there are all versions of documentation

Users can click bottom-left menu then find all versions;
image

Issue #306, opened 2021-07-11
Consideration on command line interface


miurahr:
There are many variation of command line arguments now. It is good time to rethink a design. 

for examples,

```
aqt install --outputdir /opt/Qt 5.12.0 linux desktop -m qtcharts qtnetworkauth
aqt install 5.15.2 linux android android_armv7
aqt install 5.15.0 linux desktop wasm_32

aqt examples 5.15.2 windows desktop -m qtcharts qtnetworkauth
aqt doc 5.15.2 windows desktop -m qtcharts qtnetworkauth
aqt src 5.15.2 windows desktop --archives qtbase --kde

aqt list qt5 linux desktop
aqt list qt6 mac desktop
aqt list qt6 mac desktop --modules latest
aqt list qt6 windows desktop --arch 6.1.2
aqt list tools windows desktop 
aqt list tools linux desktop --tool tools_ifw 
```

#299 introduces following cli

```
aqt tool windows desktop tools_openssl_x64
aqt tool windows android tools_qt3dstudio_runtime_240
aqt tool linux desktop tools_ifw
aqt tool mac desktop tools_ninja
```

If there is more consistency, in words such as `tools` and `tool` and common order of arguments, it would be better.

ddalcino:
We could set the first two positional arguments of each command to be `<host> <target>`. That would turn the example list into:

```
aqt install linux desktop --outputdir /opt/Qt 5.12.0 -m qtcharts qtnetworkauth
aqt install linux android 5.15.2 android_armv7
aqt install linux desktop 5.15.0 wasm_32

aqt examples windows desktop 5.15.2 -m qtcharts qtnetworkauth
aqt doc windows desktop 5.15.2 -m qtcharts qtnetworkauth
aqt src windows desktop 5.15.2 --archives qtbase --kde

aqt list linux desktop qt5
aqt list mac desktop qt6
aqt list mac desktop qt6 --modules latest
aqt list windows desktop qt6 --arch 6.1.2
aqt list windows desktop tools
aqt list linux desktop tools --tool tools_ifw 

aqt tool windows desktop tools_openssl_x64
aqt tool windows android tools_qt3dstudio_runtime_240
aqt tool linux desktop tools_ifw
aqt tool mac desktop tools_ninja

```

ddalcino:
Additionally, I think the syntax of `aqt list` needs some work. Consider these commands:

```bash
aqt list linux desktop tools                          # list all tools for linux desktop
[...]

PR #331, opened 2021-07-22
List syntax update

ddalcino:
This implements the splitting of the `aqt list` command into `aqt list-qt` and `aqt list-tool`, as discussed previously in #306. It does not address the other changes mentioned in #306; I think that's best left to another PR.

#306 mentions retaining the old syntax for the "install", "tool", "doc", and "src" commands to avoid breaking backwards compatibility. I have made no attempt to retain backwards compatibility for `aqt list`, because the feature is so new that I don't think anyone really depends on it yet, and I think it would make the code really messy.


aqt/metadata.py
```diff
@@ -41,6 +41,21 @@
...
+            "https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.SimpleSpec\n"
+            "Examples:\n"
+            '* "*": matches everything\n'
+            '* "5": matches every version with major=5\n'
+            '* "5.6": matches every version beginning with 5.6\n'
+            '* "5.*.3": matches versions with major=5 and patch=3\n'
+            '* "<5": matches every version with major<5\n'
+            '* ">5,<6.2": matches versions starting with 6.0 and 6.1\n'
+            '* "<=5": matches major<=5, including all preview/prerelease versions'
+        )
```
ddalcino, line 57: We haven't discussed this at all, but `semantic_version` also provides NpmSpec as an alternative to SimpleSpec. Their docs at https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.SimpleSpec mention an extension for PyPI-style specs as well, but I haven't been able to locate it yet.

The reason I mention this is that if we use a spec standard that everyone else already knows about, we won't need this `usage()` method to tell them how to use it. I could be very wrong about this, but I think that more of our users will know about npm specs than will know how to use SimpleSpec. I would rather not try to teach people how to use SimpleSpec if there's an alternative that they already know.

tests/test_list.py
```diff
@@ -25,51 +26,53 @@
...
+def test_versions(
+    init_data, expect_str, expect_fmt, expect_flat, expect_last, expect_bool
+):
+    versions = Versions(init_data)
+    assert str(versions) == expect_str
+    assert format(versions) == expect_fmt
+    assert format(versions, "s") == expect_str
+    assert versions.flattened() == expect_flat
+    assert versions.latest() == expect_last
+    assert bool(versions) == expect_bool
```
ddalcino, line 70: This refactor has nothing to do with `list-qt` and `list-tool`, and probably belongs in a separate PR. I can spin it off if necessary.

aqt/metadata.py
miurahr: IMO, we can drop description following example.

```
            '* "<5": matches every version with major<5\n'
            '* ">5,<6.2": matches versions starting with 6.0 and 6.1\n'
            '* "<=5": matches major<=5, including all preview/prerelease versions'
```

These are mostly a description of simple spec.

tests/test_list.py
miurahr: It is OK.

aqt/metadata.py
miurahr: It is no harm but IMHO many users may not use these complex expressions.

ddalcino: Ok, I removed the last 3; if users need more help they can read the docs.

PR #334, opened 2021-07-23
behaviour change - fail install if modules are missing

WillBrennan:
This PR changes the behaviour of aqtinstall so it will fail if a module isn't available for the requested qt version or if a module is typed incorrectly in the command line. The current behaviour is to log a warning message to the console and continue regardless.

The current behaviour can cause confusion for the end-users who may believe its a problem with their build system not being able to detect the Qt module as its easy to miss this message in the build log. This change in behaviour makes it clear that the issue was with the Qt install.

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -249,7 +249,12 @@ class Cli:
             )
         all_extra = True if modules is not None and "all" in modules else False
         if not all_extra and not self._check_modules_arg(qt_version, modules):
-            self.logger.warning("Some of specified modules are unknown.")
+            available = Settings.available_modules(qt_version)
+            self.logger.error("Some of specified modules are unknown!")
+            self.logger.error("Supported packages: {}".format(available))
+            self.logger.error("Requested packages: {}".format(modules))
+            exit()
+
         try:
             qt_archives = QtArchives(
                 os_name,


miurahr:
I'd like to wait #318 before merge it.

WillBrennan:
Sounds good. I was wondering if the combinations.json is really needed? 

For the main install script you could have the separate options for host os name and target sdk + the user enters a string for the version like they do now. You could then fetch the XML and parse the modules from that? If the XML doesn't exist then you know the version is wrong? I haven't looked at the tool arg combination to know if a similar trick can be performed? Wondering what your thoughts are?

miurahr:
There are so many issues filed but is just a typo of versions, module names in early days.
It is why output warnings.
 
And Qt company frequently change the download site contents, deprecation of versions, additions of new items. So I just show warning, rather than stop as error.

>  I was wondering if the `combinations.json` is really needed?

@ddalcino do you have any opinion for that? 
#318 is to update the `combinations.json`

ddalcino:
>>I was wondering if the `combinations.json` is really needed?

>@ddalcino do you have any opinion for that?


I would be disappointed, because it was fun to write the bot that generates the `combinations.json` file, but I agree that it would probably be better for the project to rely on the `MetadataFactory` to tell `aqt` what's in the repo and what isn't. This is definitely a job that `MetadataFactory` can do, and you could potentially use it to tell the user what was wrong with their command and how to fix it, instead of just printing a warning.

The `combinations.json` file will always go out of date eventually, and unless our users continually update `aqt`, it will probably be out of date most of the time. If `aqt` fetches data from the repo, instead of reading a cached file, `aqt`'s output won't ever go out of date until the Qt people start making large changes to the repo.

miurahr:
> The combinations.json file will always go out of date eventually, and unless our users continually update aqt, it will probably be out of date most of the time. Keep in mind, many of our users (I am one of them) use aqt through the https://github.com/jurplel/install-qt-action, which is locked to an old version

Exactly many of our users use for CI process and need some stability. @jurplel provide an option to specify the version, and also keep its stability to stick to some version.

I've been keep its behavior as warning for unknown modules, because the combinations.json file will always go out of date eventually.

 #334 compare names with combinations.json. It means that  users of some aqtinstall version  force to update when installing newer Qt versions and modules than aqtinstall released. (Or put  issues said not working...)

It may not be happy for both users and developers.

Issue #336, opened 2021-07-24
Ctrl-C during install does not work on windows


nikitalita:
On linux, ctrl-c works during install, albeit with a fatal error:
```
KeyboardInterrupt
Exception in thread Thread-1:
Fatal Python error: could not acquire lock for <_io.BufferedWriter name='<stderr>'> at interpreter shutdown, possibly due to daemon threads
Python runtime state: finalizing (tstate=0xcbb520)

Thread 0x00007fa8ad7f5700 (most recent call first):
  File "/usr/lib/python3.8/threading.py", line 1202 in invoke_excepthook
  File "/usr/lib/python3.8/threading.py", line 934 in _bootstrap_inner
  File "/usr/lib/python3.8/threading.py", line 890 in _bootstrap

Current thread 0x00007fa8b21eb740 (most recent call first):
<no Python frame>
Aborted
```

Due to the way that aqt uses multiprocessing to start multiple processes on windows (i.e. no fork, just starting multiple instances) means that ctrl-c does not terminate all the processes on windows, and results in aqt hanging after a certain point in the downloads. The processes respawn after you end them, so you will have to kill all of them at the same time in order to terminate them.

I recommend implementing something like this solution here to fix this:
https://stackoverflow.com/questions/11312525/catch-ctrlc-sigint-and-exit-multiprocesses-gracefully-in-python

miurahr:
@nikitalita does #337 works for you?

miurahr:
#337 is merged.

PR #337, opened 2021-07-24
Gracefully terminate when keyboard interrupt

miurahr:
Implement #336

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -487,7 +487,9 @@ class Cli:
                     exit(1)
             except ArchiveDownloadError or ArchiveListError:
                 exit(1)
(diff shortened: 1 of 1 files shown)


aqt/installer.py
```diff
@@ -826,11 +827,18 @@ def run_installer(
     tasks = []
     for arc in archives:
         tasks.append((arc, base_dir, sevenzip, queue, keep))
+    original_sh = signal.signal(signal.SIGINT, signal.SIG_IGN)
     ctx = multiprocessing.get_context("spawn")
     pool = ctx.Pool(Settings.concurrency)
```
miurahr, line 832: Change signal behavior before calling `Pool` is problematic because we cannot stop  when stack.

Issue #348, opened 2021-07-27
Question: Can aqt install qtcreator?


billpriest:
I was trying to install the latest qtcreator w/o the "online/offline" installer both of which have issues w/ the platform plugin.

ddalcino:
The prerelease versions of aqt (v2.0.0b5) can install qtcreator using this command:

``` bash
    $ aqt tool <host> <target> tools_qtcreator
```

The stable release of aqt (v1.2.4) has an older version of the `aqt tool` subcommand; using it will require that you read some Updates.xml files at https://download.qt.io/online/qtsdkrepository/ to find out the version and PackageUpdate that you want to install; I do not recommend this.

However, you mentioned before that you were running debian buster on a Chromebook. In my experience, Debian runs in a VM on Chromebooks, and it cannot launch GUI apps. Be aware that this might be a problem.

billpriest:
Hi David,
  Thanks for the info.  I'll give it a shot.  It seems crostini has come a long way, I can run all kinds of GUI apps (gitkraken, emacs, qtcreator, etc.) w/o any serious issue.  There are some visual quirks in qtcreator (pop-up windows that don't stay in the viewable area if you click too near the top of the editor window; I mostly use shortcuts so it is doesn't bother me much).  There is an issue using the debugger where I try to click on a set breakpoint that it pops up the complete list of breakpoints which I really dislike (tooltips on steroids).  I'll provide feedback.  What is it looking for <target>?

billpriest:
In case someone else tries qtcreator on debian, there is a missing liib libxcb-util.so.1 which can be fixed by making a sym link to libxcb-util.so.  QtCreator 4.15.2 works like a charm w/o the obnoxious "breakpoint popup".

Issue #339, opened 2021-07-25
Documentation improvements


ddalcino:
I have several different ideas for improvements to the documentation, and I'd like to discuss them here.

### 1. More consistent names for CLI arguments.
For instance, the help text for most subcommands refers to the 'host operating system' (windows, mac, linux) as 'host'. However, the documentation refers to it as 'target OS', and the source code alternates between calling it 'host' and 'os_name'. I think it's a mistake to call it 'target OS' anywhere, because the docs refer to many CLI arguments as 'target something', and the help text refers to the next positional argument as 'target' (desktop, android, ios, winrt). There are several instances of inconsistent or confusing names that I would like to change:

* 'target OS'/'host'/'os_name' (windows, mac, linux):
I would like to call this 'host' in all places.
* 'target variant'/'target'/'target_sdk' (desktop, android, ios, winrt): 
I would like to call this 'target' in all places.
* 'target architecture'/'arch'/'architecture' (gcc_64 , clang_64, win_msvc, mingw, android_armv7, etc):
I would like to call this 'arch' in all places.
* 'tool variant name'/'arch' (qt.tools.openssl.gcc_64, qt.tools.vcredist_msvc2013_x64, qt.tools.ifw.41):
I would like to call this 'tool_variant' or 'tool_module' in all places. This argument is referred to as 'arch' in the `aqt tool` help text and the `ToolArchive` code. The names sometimes include an architecture, but they are definitely not architectures. Calling this argument 'arch' is confusing, and it clashes with the other 'arch' argument.

Honestly, I don't care what we change these names to, as long as they are consistent in the help text, code, and docs.

### 2. Subcommand formatting

When I split the `aqt list` command into `list-qt` and `list-tool`, I reformatted the docs to list a single command line option on every line, so they would be more readable. When displayed as code, they look ok:
```
.. program::  aqt

.. option:: list-qt [-h | --help]
                    [--extension <extension>]
                    [--spec <specification>]
                    [--modules    (<Qt version> | latest) |
                     --extensions (<Qt version> | latest) |
                     --arch       (<Qt version> | latest) |
                     --latest-version]
                    <target OS> [<target variant>]

    List available versions of Qt, targets, extensions, modules, and architectures.
```
However, Sphinx formats this code into this mess:

![image](https://user-images.githubusercontent.com/6743441/126909468-0a050f77-2fd2-4b15-a897-d5a51200bffb.png)

[...]

Issue #333, opened 2021-07-23
Issue with QtCharts in 6.2.0


WillBrennan:
Hi there; its a really great tool you've made!

I've just been trying to set up CI with github actions for a personal project. But I was having issues installing Qt with,

```bash
aqt install 6.2.0 --outputdir ./qt linux desktop -m qtcharts
```
. When I ran it locally I could see that its missing; 

- QtCharts from the include directory
- Qt6Charts in lib/cmake
- Charts.json from the modules directory
- and the libQt6Charts.so library

. Given these missing files and that my CMakeLists works on my local install of Qt6 installed with the Qt installer I'm pretty sure its failed to install the optional module. It has managed to install the standard Core + Widgets libraries however. 

I'm going to dig a little bit deeper and to try and figure out why.

WillBrennan:
It appears that they're missing in this configuration JSON you've got set here, 

https://github.com/miurahr/aqtinstall/blame/08b66255cb184e132112350601b1547f9f4e6910/aqt/combinations.json#L78

. I can't see any comments on why QtCharts wasn't added. I'll try adding them and if it works I'll make a PR.

WillBrennan:
(just some personal notes); looks like its more than just adding it to the combinations.json, looks like it gets filtered in the `QtArchives` class and it can't resolve the package.

miurahr:
@WillBrennan could you put log file when you tried?

WillBrennan:
Hey - thanks for the quick response; 

It says some of the modules are unknown from this logging statement, [installer.py#L252](https://github.com/miurahr/aqtinstall/blob/master/aqt/installer.py#L252) where it couldn't find `QtCharts` in the combinations.json file.

When you add it to the modules section of the combinations.json file it gets filtered in the constructor of the QtArchive class. The module gets discovered and passed into the class, but then its package doesn't appear in `QtArchive.get_packages` like it does on Qt5. I'm just looking at the XML files from Qt now and seeing whats different between them.

```
2021-07-23 16:02:45,449 - aqt.main - INFO - installer 140492440684352 aqtinstall(aqt) v1.2.4 on Python 3.7.4 [CPython GCC 7.3.0]
2021-07-23 16:02:45,449 - aqt.main - WARNING - installer 140492440684352 Some of specified modules are unknown.
2021-07-23 16:02:46,157 - aqt.installer - INFO - installer 140306725533504 Downloading qtbase...
[...]

PR #351, opened 2021-07-27
Add Getting Started section

ddalcino:
This adds a tutorial that should prevent issues like #348, #333, and many others.

Partial fix for #339


ddalcino:
I think this is done. I will wait until #306 is closed to remove the 'WIP' label though, because the tutorial uses the new command syntax for `aqt install-qt` and `aqt install-tool`, and I don't want new users to try to use the new commands before they exist.

miurahr:
>  I will wait until #306 is closed

Oh, I am a blocker now. 😸

PR #359, opened 2021-08-03
Add 'tests/test_install.py'

ddalcino:
This change adds unit tests for the `aqt install` and `aqt install-qt` commands, and establishes a pattern that can be extended for more tests. This is intended to make it easier to increase test coverage of parts of the codebase that are not yet covered by tests.

This uses the `pytest-socket` library to ensure that the tests do not use any network IO. It also mocks `multiprocessing.get_context` to prevent multiprocessing. This is necessary because multiprocessing spawns child processes that have not been monkey-patched, which would break the test.

These tests use py7zr to create mock 7z archives, which are installed and patched in a temporary directory. The tests check the content of the patched files and any output to stderr.

I think that this will increase the code coverage of this project by a good chunk. These are the coverage changes that I am seeing locally:

* archives: 56 -> 59%
* helper: 66 -> 74%
* installer: 42 -> 65%
* updater: 11 -> 53%

**Edit:** The tests are running OK locally. I'm not yet certain why they are failing in CI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,6 +57,7 @@ test =
     pytest-pep8
     pytest-remotedata
     pytest-cov
+    pytest-socket
 check =
     flake8
     flake8-black
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -249,7 +249,7 @@ def test_install(
     arch_dir: str,
     updates_url: str,
     files: Iterable[Dict[str, str]],
-    expect_out: re.Pattern,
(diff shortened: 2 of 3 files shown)


ddalcino:
This definitely needs a code review. I did some weird things to monkeypatch the multiprocessing library that I am not too comfortable with. I have noticed some platform-dependent differences between Windows and Linux, and that makes me wonder if these tests will run properly on a Mac. I have not tried that yet, since I don't have regular access to a Mac.

PR #358, opened 2021-08-03
Update docs

ddalcino:
This should address any remaining issues brought up in #339.

This PR does a lot of different things; see the individual commit messages for explanations.

In summary, it does the following:
1. fixes a broken table,
2. fixes broken links
3. adds helpful links
4. adds discussion of previously undocumented flags
5. replaces any mention of `target OS` and `target variant` with `host` and `target`, to be more consistent with help text and source code
6. cleanup of language in several places

Fixes #339

PR #360, opened 2021-08-04
Turn warnings into errors when building docs

ddalcino:
This turns on 'warnings as errors' for `sphinx-build`, which will help to prevent problems in the documentation. 

The CI for this PR will fail because the master branch contains broken links in `docs/getting_started.rst`, for the 'Available Versions' wiki page. #358 will fix these broken links.

Here's an example run where the build failed because of a broken link in the documentation:
https://github.com/ddalcino/aqtinstall/runs/3245370308

Here's a run where the link was fixed, causing the build to succeed:
https://github.com/ddalcino/aqtinstall/runs/3245406178

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -37,8 +37,8 @@ commands =
 basepython = python3.8
 extras = docs
 commands =
-    sphinx-build {posargs:-E} -b html docs build/docs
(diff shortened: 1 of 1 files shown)


miurahr:
master updated to do it.

PR #366, opened 2021-08-14
Fix change log formatting

ddalcino:
This fixes the formatting for the last 3 releases, which causes the links/table of contents entries to work properly.

This also fixes the date for v1.2.5.


ddalcino:
Also, it looks like the change log entries for the last 3 stable releases haven’t gotten into the change log for the master branch yet.

miurahr:
Thank you!

PR #335, opened 2021-07-23
add support for qt6 optional modules

WillBrennan:
Currently Qt6 fails to install optional modules such as Qt::Charts or Qt::WebViewer. This is because the syntax in the qt sdk repository has changed. In Qt 6 onwards the package names have changed from qt.qt5.5150.qtcharts.clang_64 to qt.qt6.620.addons.qtcharts.clang_64. The addition of this "addon" keyword broke the XML parser in the QtArchives class. 

https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt6_620/Updates.xml
https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt5_5150/Updates.xml

You can see it now appears correctly in the logs, 
```
aqtinstall(aqt) v2.0.0b4.dev66 on Python 3.8.5 [CPython Clang 10.0.0 ]
Downloading qtbase...
Downloading qtcharts...
Downloading qtsvg...
Downloading qtdeclarative...
Redirected: mirrors.ukfast.co.uk
Redirected: mirrors.ukfast.co.uk
Redirected: mirrors.ukfast.co.uk
Redirected: mirrors.ukfast.co.uk
Finished installation of qtsvg-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 0.78178350
Downloading qttools...
Finished installation of qtcharts-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 1.30998046
Downloading qtquickcontrols2...
Redirected: mirrors.ukfast.co.uk
Redirected: mirrors.ukfast.co.uk
Finished installation of qtquickcontrols2-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 5.61572662
Downloading qttranslations...
Redirected: mirrors.ukfast.co.uk
Finished installation of qttranslations-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 3.71476571
Finished installation of qtbase-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 12.46363146
Finished installation of qttools-MacOS-MacOS_11_00-Clang-MacOS-MacOS_11_00-X86_64-ARM64.7z in 31.24148654
[...]

PR #368, opened 2021-08-18
Remove `addons.` prefix from Qt 6.1/6.2 module names

### LINKED ISSUE #333: Issue with QtCharts in 6.2.0

ddalcino:
This implements the first change suggested in https://github.com/miurahr/aqtinstall/discussions/367#discussioncomment-1197554.

I think this should fix #333.
Supersedes #335.

To do this, I had to make a large refactor to the tests in test_install.py; this was necessary so that I could add tests that check that the modules were being installed properly. I am concerned that the tests are getting a little harder to read. This is as readable as I think I can make them right now, but surely the readability can still be improved.


miurahr:
Oh, this is so small changes to realize it.

PR #375, opened 2021-08-24
Fail installation on missing tool variants

ddalcino:
This implements the other half of #374, so that `aqt install-tool` should fail if you tell it to install a tool variant that doesn't exist.


miurahr:
LGTM

PR #385, opened 2021-09-01
Improve error handling

### LINKED ISSUE #381: `Cli.run` return value / error handling strategy
Here's the code for `Cli.run`:

https://github.com/miurahr/aqtinstall/blob/179a9c68f948562e9768fceba36f8457bedd6845/aqt/installer.py#L86-L90

The return value from this function is eventually returned to the command prompt by an earlier `main` function call. By reading this code, you would expect that the set of functions that can be assigned to `args.func`, aka `run_install_src`, `run_install_doc`, `run_install_tool`, `run_install_qt`, `run_list_qt`, and `run_list_tool`, would all return a `1` or a `0`, or some other numeric value. However, they do not. The `run_install_*` functions all return `None`, which is coerced into a `0`, and whenever an error occurs, the function calls `self.logger.error(some_error_message)` followed by `exit(1)`.  Several instances of `exit(1)` are not preceded by an error message, so it's unclear whether the user is being told what the problem is, or that there even is a problem. There are multiple copies of something like this:

https://github.com/miurahr/aqtinstall/blob/179a9c68f948562e9768fceba36f8457bedd6845/aqt/installer.py#L249-L251

There's a lot of inconsistently repeated code here.

This is the design that I would have expected for `Cli.run()`:
```python
def run(self, arg=None):
  args = self.parser.parse_args(arg)
  self._setup_settings(args)
  try:
    args.func(args)
    return 0
  except SomeExpectedBaseException as e:
    self.logger.error(format(e))
    return 1
```

All of the `run_install_*` or `run_list_*` functions would return `None`. If any errors occur that cannot be recovered from, they would raise an exception that contains a useful error message, so that `Cli.run` can log it. None of these functions would attempt to catch exceptions unless they are recoverable, or the error message needs to be modified in some way. Additionally, any time an exception is raised elsewhere in the program, that exception would contain whatever error message is appropriate to show the user what went wrong, instead of logging the message and raising an empty exception.

There's one disadvantage of this approach: Currently, when exceptions are raised, the code that raises the exception also logs the error message using its own logger, which means the name of the logger is attached to the log message, so you can tell what file it's coming from. If that data is really important, we could use `self.logger.exception(e)` in `Cli.run` instead of `self.logger.error(e)`; this would attach a stack trace to the log, which would include filenames and line numbers.

Please let me know if you think this kind of thing would be an improvement. I think it would be relatively easy to implement this change, it would remove some repeated code, and I think it would make the project easier to maintain. However, that's just my opinion; this kind of refactor is very subjective, and I totally understand if you think I'm trying to fix something that isn't really broken. For that reason, I won't start on this unless you think it's a good idea.

ddalcino:
This migrates the error handling strategy of `aqt/installer.py` from "print error message, then exit(1)" towards consistently raising and catching exceptions. See #381 for more details.

This also adds an option to `settings.ini` that allows you to turn stack traces on and off during error reporting. This behavior could probably be triggered by a command line option, but I didn't think of that.

Fix #381.

**Note**: I intend to keep this on WIP until I have test coverage for every instance of `raise AqtException` that I have altered or added.

**Edit**: There are now at least 11 instances of `raise AqtException` that I have altered or added:

- [x] 2 uncovered exceptions in archives.py
- [x] 5 uncovered exceptions in helper.py
- [x] 3 uncovered exceptions in installer.py
- [x] 1 uncovered exception in updater.py

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -262,7 +262,6 @@ class QtArchives:
         # if we have located every requested package, then target_packages will be empty
         if len(target_packages) > 0:
             message = f"The packages {target_packages} were not found while parsing XML of package information!"
-            self.logger.error(message)
             raise NoPackageFound(message)
 
     def get_packages(self) -> List[QtPackage]:
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -212,8 +212,7 @@ class QtArchives:
         try:
             self.update_xml = ElementTree.fromstring(self.update_xml_text)
         except ElementTree.ParseError as perror:
-            self.logger.error("Downloaded metadata is corrupted. {}".format(perror))
-            raise ArchiveListError("Downloaded metadata is corrupted.")
+            raise ArchiveListError(f"Downloaded metadata is corrupted. {perror}") from perror
 
         for packageupdate in self.update_xml.iter("PackageUpdate"):
             pkg_name = packageupdate.find("Name").text
@@ -387,8 +386,7 @@ class ToolArchives(QtArchives):
         try:
             self.update_xml = ElementTree.fromstring(self.update_xml_text)
         except ElementTree.ParseError as perror:
-            self.logger.error("Downloaded metadata is corrupted. {}".format(perror))
-            raise ArchiveListError("Downloaded metadata is corrupted.")
+            raise ArchiveListError(f"Downloaded metadata is corrupted. {perror}") from perror
 
         try:
             packageupdate = next(filter(lambda x: x.find("Name").text == self.arch, self.update_xml.iter("PackageUpdate")))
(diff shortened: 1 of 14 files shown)


ddalcino:
> Good direction and great improvement!

Thank you for the encouraging words! This is a much bigger refactor than I meant it to be, and I am relieved that my efforts are appreciated!

ddalcino:
Ok, this is ready for review. Again, I'm sorry this is such a huge change; please let me know if you would like me to split it up in some way. I did not feel comfortable making this change without adding a lot more test code.

In `aqt/installer.py`, there are still 6 lines that I added that are still not covered, and 1 existing line that is no longer covered (I don't know what happened there). You can see these lines here: https://coveralls.io/builds/42686339.

Please let me know if you'd like me to add tests to cover those 7 lines.

miurahr:
LGTM.
I'd like to merge it now.
I don't find a point for further changes, and I'd like to ask users for feedback from releases.

PR #389, opened 2021-09-03
Add readthedocs config yaml and use version in conf.py

### LINKED ISSUE #380: Docs: remember to update the version number in conf.py
Hi.

On https://aqtinstall.readthedocs.io/en/latest/ , I got "aqtinstall 1.0.0b2 documentation" from its title; in the PDF file downloaded from https://aqtinstall.readthedocs.io/_/downloads/en/latest/pdf/ , I got "Release 1.0.0b2" from its cover and header. The version number of "1.0.0b2" does not match the version I selected.

This version number is defined in /docs/conf.py. It seems that it has not been updated for a long time:

https://github.com/miurahr/aqtinstall/blob/4f0d92dc977ca9255f7ac912e1fd4fc303e953c1/docs/conf.py#L59-L67

I think we need to update it before releasing each version, otherwise it may confuse the user.

The docs of Sphinx: https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information

miurahr:
resolve #380


pzhlkj6612:
OK, I didn't know ".readthedocs.yml". It looks more convenient than manually building and publishing documents by some CI steps.

PR #390, opened 2021-09-06
In the "check combinations.json" workflow, update the "update-combinations" branch as soon as possible to avoid merge conflicts

pzhlkj6612:
See #386 .

Q: Why update so early?
A: The ["check combinations.json" step](https://github.com/miurahr/aqtinstall/blob/c094930a489795cedb11124fff89511825eb31d5/.github/workflows/check-combinations.yml#L34-L35) will update the "/aqt/combinations.json" file. If I check out the "update-combinations" branch after this step, git will complain that our *local changes would be overwritten by merge*. Also, I think it's not a bad thing to put git stuff together.

Q: Why use `branchtomerge: "origin/master"` , not `branchtomerge: "master"` ?
A: At that time, there was no "master" branch locally.

/cc @ddalcino

### DIFF
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -17,6 +17,14 @@ jobs:
     - uses: actions/checkout@master
       with:
         fetch-depth: 20
+      
+    - name: Update the "update-combinations" branch (ff merge to master)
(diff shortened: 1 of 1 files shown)


ddalcino (approved):
LGTM

PR #251, opened 2021-05-24
WIP: feature to use offline installer

miurahr:
Trial to integrate CuteCI installation logics into aqtinstall.
This is incomplete.

There will be no further work without any  contribution from one who have a knowledge and experiments about offline_installer.

Issue #236, opened 2021-05-05
Add support for old qt versions (<5.9) (Trying to use 5.6)


PibePlayer:
I'm trying to use this for a github action that requires QT 5.6, but it isn't on 
[QtSDKRepo](https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/) unless I use the 56_preview, which fails with
```
`2021-05-05 17:27:46,964 - aqt - INFO - aqtinstall(aqt.installer) vunknown on Python 3.9.4 [CPython MSC v.1928 64 bit (AMD64)]
2021-05-05 17:27:46,964 - aqt - WARNING - Specified Qt version is unknown: 5.6_preview.
2021-05-05 17:27:46,965 - aqt - WARNING - Specified target combination is not valid or unknown: windows desktop win32_mingw49
2021-05-05 17:27:47,755 - aqt - ERROR - Specified packages are not found while parsing XML of package information!
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\__main__.py", line 27, in <module>
    sys.exit(main())
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\__init__.py", line 38, in main
    return cli.run()
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\installer.py", line 471, in run
    return args.func(args)
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\installer.py", line 220, in run_install
    qt_archives = QtArchives(os_name, target, qt_version, arch, base, subarchives=archives, modules=modules,
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\archives.py", line 159, in __init__
    self._get_archives(qt_ver_num)
  File "C:\hostedtoolcache\windows\Python\3.9.4\x64\lib\site-packages\aqt\archives.py", line 182, in _get_archives
    self._parse_update_xml(archive_url, target_packages)
[...]

PR #393, opened 2021-09-08
Add Windows standalone binary

nikitalita:
This pull request adds:

- a script for building a standalone binary on Windows
- minor code changes to accommodate running `aqt` as a standalone binary
- workflow for building and testing the standalone binary on windows
- workflow for automatically building and uploading the standalone binary when a new release is created


miurahr:
There seems many code duplication among .github/workflows/build-standalone.yml with other existent script.(You may copy)

Could you manage a standalone executable generation and test with  github actions matrix variations?

Here is an just idea but we can put  matrix variable `artifact: [standard, binary]` and switch on/off with actions `if:` clause and/or python if condition in the script. You can put `binary` only for windows using exclude/include clause of actions.

miurahr:
> The reason I added this is because there are a number of setup scripts for build environments on Windows which require QT.

Python is installed on most CI cloud services, but when you need to setup your own build servers, it will be cumbersome.

KOLANICH:
IMHO a standalone binary is not needed. Python would be needed by end users anyway, sooner or later since lot of tools now depend on python, including build tools such as meson. So end users should just install the distro of python that seems to be matching their needs better and not to install bloated bundled unupdateable versions of python.

nikitalita:
> IMHO a standalone binary with bundled python is not needed. Python would be needed by end users anyway, sooner or later since lot of tools now depend on python, including build tools such as meson. So end users should just install the distro of python that seems to be matching their needs better and not to install bloated bundled unupdateable versions of python.

There are a number of problems with this:

1) On Windows, actually locating Python is not straightforward; you don't actually know that python will be on the PATH and not in some god-forsaken Conda environment or what-have-you, 
2) PIP's approach to dependency management means that installing ANY package with a frequently used dependency could potentially break your environment
3) It's not strictly necessary, and for a bespoke script for installing a QT version where aqtinstall may be a temporary tool, installing Python too is overkill.

miurahr:
> There are a number of problems with this:

IMHO, It seems that the genuine Qt installer is best for you. It doesn't request a python, single binary, and works well with Windows GUI environment. It also provides an automation mechanism through Qt scripting (but I don't know well). 

This abandoned PR (https://github.com/miurahr/aqtinstall/pull/251/files) and discussion on a closed issue (https://github.com/miurahr/aqtinstall/issues/236) might give you some hint (see install-qt.qs file in PR).

Even though my above comment, I'd like to add a binary version of aqtinstall distribution,  @nikitalita  suggested, on github release, because it seems not harm for another artifacts, and @nikitalita suggest a release automation on github actions.

KOLANICH:
>On Windows, actually locating Python is not straightforward; you don't actually know that python will be on the PATH and not in some god-forsaken Conda environment or what-have-you,

Easily solved by installing everything into the global environment and adding the binary from conda distro into `%PATH%`. If some program bundles own python, it is their problems to solve.

>PIP's approach to dependency management means that installing ANY package with a frequently used dependency could potentially break your environment

Solved by updating all the broken packages to their latest versions, where incompatibilities are fixed. If they are not fixed, then it is not the latest version and the latest is the one with a patch fixing the incompatibilities landed/rebased over it :)

>aqtinstall may be a temporary tool, installing Python too is overkill.

One still have to install it ... most of software is installed by just unpacking it. One still has to unpack code in order to run it locally (or one can use cloud where everything is installed and unpacked by cloud owner).

nikitalita:
> There seems many code duplication among .github/workflows/build-standalone.yml with other existent script.(You may copy)
> 
> Could you manage a standalone executable generation and test with github actions matrix variations?
> 
> Here is an just idea but we can put matrix variable `artifact: [standard, binary]` and switch on/off with actions `if:` clause and/or python if condition in the script. You can put `binary` only for windows using exclude/include clause of actions.

Ok, this is done.

miurahr:
LGTM.

nikitalita:
> Does v2.0.0rc4 is a status the PR intended?
> It just have an exe file in download list.
> Why not explain what it is in README?

Yes, that is intended; every time a new release is created, it will build and upload a new standalone release artifact, which is just the bare exe.
Do you want it to be in a zip with a README in it instead?

ddalcino:
I think the bare `aqt.exe` is exactly what we need in the download list.

However, we definitely need to add a section to the README, probably in the installation instructions, that discusses where to find `aqt.exe` and why you might want to use it instead of installing via pip.

Additionally, I think it may be useful to offer a continuous release, like the one from [linuxdeployqt](https://github.com/probonopd/linuxdeployqt/releases/continuous). To download the most current version of linuxdeployqt, all you have to do is `curl` https://github.com/probonopd/linuxdeployqt/releases/continuous/linuxdeployqt-continuous-x86_64.AppImage. The link never changes, so you never have to update your CI code to point to newer releases.

pzhlkj6612:
> Additionally, I think it may be useful to offer a continuous release, like the one from [linuxdeployqt](https://github.com/probonopd/linuxdeployqt/releases/continuous). ...

See https://github.com/miurahr/aqtinstall/pull/397#discussion_r710540102 .

.gitignore
```diff
@@ -2,6 +2,7 @@ build
 aqt/version.py
 dist
 venv
+build_venv
```
pzhlkj6612, line 5:
```suggestion

```

Hi. I found that virtualenv will generate a ".gitignore" file in the virtual environment directory automatically, so it's unnecessary to ignore it manually.

> ## [v20.0.21 (2020-05-20)](https://virtualenv.pypa.io/en/latest/changelog.html#v20-0-21-2020-05-20)
> ### [Features - 20.0.21](https://virtualenv.pypa.io/en/latest/changelog.html#features-20-0-21)
> - Generate ignore file for version control systems to avoid tracking virtual environments by default. Users should remove these files if still want to track. For now we support only **git** by **gaborbernat**. ([#1806](https://github.com/pypa/virtualenv/issues/1806))

PR #400, opened 2021-09-17
Add `list-qt --archives` feature

### LINKED ISSUE #394: Allow `list-qt` to print a list of archives for use with the `--archives` option
Let's say you want to install Qt 5.15.0 for linux desktop, but only install the 7z archives for {qtbase, qtdeclarative, icu}. You would use this command:

```
aqt install-qt linux desktop 5.15.0 --archives qtbase qtdeclarative icu
```

This feature is documented  at https://aqtinstall.readthedocs.io/en/latest/cli.html#cmdoption-list-tool-archives, and there's an example under https://aqtinstall.readthedocs.io/en/latest/cli.html#command-examples; it's the third example in the list. 

I think it would be useful to add an option that lists archives to `list-qt`, but I cannot do it without an idea of what the interface would look like. I would like to use this issue to figure out what that interface would look like.

_Originally posted by @ddalcino in https://github.com/miurahr/aqtinstall/issues/382#issuecomment-911653563_

ddalcino:
Fix #394


ddalcino:
@pzhlkj6612, can you please review my changes to the documentation? I tried my best to explain the new feature, but I'm not confident that my documentation is easy enough to understand.

docs/getting_started.rst
```diff
@@ -405,3 +405,127 @@ the installer may use the ``installscript.qs`` script to make additional changes
...
+to print which archives are part of the ``debug_info`` module:
+
+.. code-block:: console
+
+    $ aqt list-qt linux desktop --archives 5.15.2 gcc_64 debug_info
+    qt3d qtbase qtcharts qtconnectivity qtdatavis3d qtdeclarative qtgamepad qtgraphicaleffects
+    qtimageformats qtlocation qtlottie qtmultimedia qtnetworkauth qtpurchasing qtquick3d
+    qtquickcontrols qtquickcontrols2 qtquicktimeline qtremoteobjects qtscript qtscxml qtsensors
+    qtserialbus qtserialport qtspeech qtsvg qttools qtvirtualkeyboard qtwayland qtwebchannel
+    qtwebengine qtwebglplugin qtwebsockets qtwebview qtx11extras qtxmlpatterns
```
ddalcino, line 490: There's a name collision between the `debug_info` archives and the archives in every other module/Qt base install. As I understand it, if you install `debug_info` and `qtcharts`, and specify `aqt install-qt --archives qtcharts`, you will install the `qtcharts` archives from both the `qtcharts` module and the `debug_info` module.

I think this name collision makes the feature harder to understand. I tried to explain it below, but I'm sure this is good enough.

miurahr:
Tests failed because of coveralls.io outage.

pzhlkj6612:
@ddalcino 

Hmm, I‘ve been lost in the documentation (248329c4572be1aff4ea36edf80deb4467783243) because it's a bit complicated. I'll try to make it better, and it may take a lot of time.

The only suggestion I can provide right now is to divide the content of "Installing a subset of Qt archives [Advanced]" into sections (e.g., "introduction", "installing qtbase only", "installing more archives", "installing debug info", etc). I think this would make the content clearer.

ddalcino:
> The only suggestion I can provide right now is to divide the content of "Installing a subset of Qt archives [Advanced]" into sections (e.g., "introduction", "installing qtbase only", "installing more archives", "installing debug info", etc). I think this would make the content clearer.

That’s a great idea; I’m glad I asked you for help! I’ll take another look at this with fresh eyes next week. We should be able to fix this.

ddalcino:
The CI is failing because the mirror link, https://mirrors.ocf.berkeley.edu/qt/, is down right now. Sphinx is checking every link on the page to see if they are broken.

Is there another link we can use instead of the Berkeley one? I don't want to turn off the link checking behavior; it's really useful for the internal links.

miurahr:
All without mirror site is OK and I'd like to merge now.

Issue #371, opened 2021-08-18
List-qt interface change may be required to fix misleading output


ddalcino:
I just noticed that there appear to be some modules that exist for some architectures, but not all of them. `aqt list-qt --modules` does not take architecture into account when it lists all available modules: it just prints all the modules available, regardless of architecture.

Here's a concrete example:

```console
$ python -m aqt list-qt windows desktop --modules 5.14.0
debug_info qtcharts qtdatavis3d qtlottie qtnetworkauth qtpurchasing qtquick3d qtquicktimeline qtscript 
qtvirtualkeyboard qtwebengine qtwebglplugin
```

All of these modules are available when you're using an MSVC architecture, but `debug_info` is not available if you're using mingw.

I don't know if this is a big deal or not. I think that this is a very low-priority issue, and I only opened it so that I don't forget about it later.

Maybe we can just make a note in the docs and close this issue. If it turns out this could be a real problem, we could change the `--modules` interface to require an architecture, and then only print modules for a specific architecture.

ddalcino:
Additionally, for Qt 5.14.0 on Windows desktop, `qtwebengine` exists for the architectures `win64_msvc2017_64` and `win32_msvc2017`, but not mingw or `win64_msvc2015_64`.

ddalcino:
I think the most correct way to solve this issue is to redesign the `aqt list-qt` interface, so that a request for available modules requires that the user enters an architecture, if there are multiple architectures available.

Here's the existing interface:
```console
$ aqt list-qt mac                                            # print all targets for Mac OS
$ aqt list-qt mac desktop                                    # print all versions of Qt
$ aqt list-qt mac desktop --extension wasm                   # print all wasm versions of Qt
$ aqt list-qt mac desktop --spec "5.9"                       # print all versions of Qt 5.9
$ aqt list-qt mac desktop --spec "5.9" --latest-version      # print latest Qt 5.9
$ aqt list-qt mac desktop --modules 5.12.0                   # print modules for 5.12.0
$ aqt list-qt mac desktop --spec 5.9 --modules latest        # print modules for latest 5.9
$ aqt list-qt mac desktop --extensions 5.9.0                 # print choices for --extension flag
$ aqt list-qt mac desktop --arch 5.9.9                       # print architectures for 5.9.9/mac/desktop
$ aqt list-qt mac desktop --arch latest                      # print architectures for the latest Qt
```

Adding the architecture to `--modules` changes it to:
```console
[...]

PR #401, opened 2021-09-17
Require architecture when listing modules

ddalcino:
This PR solves the problem in #371 by requiring that the user enter the architecture when listing modules, like this:

```bash
$ aqt list-qt mac desktop --modules 5.12.0 clang_64                     # print modules for 5.12.0
$ aqt list-qt mac desktop --spec 5.9 --modules latest clang_64          # print modules for latest 5.9
$ aqt list-qt windows desktop --modules 5.12.0  win64_mingw73           # print modules for 5.12.0
$ aqt list-qt windows desktop --spec 5.9 --modules latest win64_mingw73 # print modules for latest 5.9
```

This solution does not attempt to automatically deduce the architecture when only one architecture is available. This feature was suggested in #371, and it's definitely possible, but I have not implemented it.

It also does not attempt to implement the improved interface proposed in #371. The improved interface is a bigger job than I expected it to be, and I think it can be done independently from this PR.


miurahr:
Thanks~!

PR #402, opened 2021-09-20
Add helpful error messages when modules/tools/Qt version does not exist

ddalcino:
This adds error messages that suggest using `aqt list-*` when a user attempts to install:
1. Versions of Qt that don't exist
2. Modules that don't exist
3. Tools that don't exist
4. Tool variants that don't exist

See changes to `tests/test_install.py` for examples of the expected error messages.

The idea here is that the user will be encouraged to use `aqt list-qt` and `aqt list-tool` when one of these errors occur, instead of trying to figure out why a 404 occurred, and trying to debug `aqt` to figure out what went wrong.

There's one thing that I didn't do that could potentially improve this: We could automatically run the suggested `aqt list-*` command and print "These are the available values for the parameter you got wrong: [...]" in the error message, instead of asking the user to run `aqt list-*`.

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -271,17 +271,16 @@ class QtArchives:
             raise NoPackageFound(message, suggested_action=self.help_msg(target_packages.get_modules()))
 
     def help_msg(self, missing_modules: Iterable[str]) -> Iterable[str]:
-        num_missing = len(list(missing_modules))
-        assert num_missing > 0
-        arch = "Please use 'aqt list-qt {0.os_name} {0.target} --arch {0.version}' to show architectures available."
-        mods = "Please use 'aqt list-qt {0.os_name} {0.target} --modules {0.version} <arch>' to show modules available."
+        base_cmd = f"aqt list-qt {self.os_name} {self.target}"
+        arch = f"Please use '{base_cmd} --arch {self.version}' to show architectures available."
+        mods = f"Please use '{base_cmd} --modules {self.version} <arch>' to show modules available."
         has_base_pkg: bool = self._base_target_package_name() in missing_modules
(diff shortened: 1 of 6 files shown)


aqt/archives.py
```diff
@@ -257,7 +268,20 @@ def _parse_update_xml(self, archive_url, target_packages: Optional[ModuleToPacka
         # if we have located every requested package, then target_packages will be empty
         if len(target_packages) > 0:
             message = f"The packages {target_packages} were not found while parsing XML of package information!"
-            raise NoPackageFound(message)
+            raise NoPackageFound(message, suggested_action=self.help_msg(target_packages.get_modules()))
+
+    def help_msg(self, missing_modules: Iterable[str]) -> Iterable[str]:
+        base_cmd = f"aqt list-qt {self.os_name} {self.target}"
+        arch = f"Please use '{base_cmd} --arch {self.version}' to show architectures available."
+        mods = f"Please use '{base_cmd} --modules {self.version} <arch>' to show modules available."
```
ddalcino, line 276: This particular error message won't make sense unless #401 is merged, because it asks the user to enter architecture as a parameter to `--modules`. I can change it if necessary, but I thought it would be easier to write it like this so we don't have to change it when #401 is merged.

**Note:** In the context of `QtArchives`, we could replace the `<arch>` term with `{self.arch}`, and the user could just copy and paste the resulting command. However, `SrcDocExamplesArchives` uses this code as well, and in `SrcDocExamplesArchives` the `self.arch` data member stores "src", "doc", or "example", which would not work as an arch in the `aqt list-qt` command. I wasn't sure about the best way to resolve this, so I left it as `<arch>`.

PR #403, opened 2021-09-25
Fix broken mirror link in cli.rst 

ddalcino:
This is meant to allow CI builds to run successfully when the mirror link in `cli.rst` is broken. It accomplishes this by changing the link into an html 'code' tag. The link is not meant to be clicked on; it is meant to be typed as a command line argument. This change makes the intention more explicit.

Additionally, this replaces the url with a working URL, in case anyone is tempted to copy it into the terminal.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -212,7 +212,7 @@ are described here:
 
 .. option:: --base, -b <base url>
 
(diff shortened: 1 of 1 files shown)


ddalcino:
On second thought, it might be useful to leave the mirror link as a link and not a code tag, so our CI will alert us when we need to update the link. I can change this if necessary.

PR #410, opened 2021-09-29
CI: update release workflow action

miurahr:
- Run on Ubuntu 20.04
- Check uploading  test.pypi.org only for releases branch
- Rename action script
- Checkout git depth to 50

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,4 +1,4 @@
-name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distributions 📦 to PyPI
 
 on:
   push:
@@ -11,13 +11,13 @@ on:
 
 jobs:
   build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
-    runs-on: ubuntu-18.04
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2
         with:
-          fetch-depth: 20
+          fetch-depth: 50
       - name: Fetch release tag
         run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
       - name: Set up Python 3.7🐍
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -11,13 +11,13 @@ on:
...
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
-    runs-on: ubuntu-18.04
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2
         with:
-          fetch-depth: 20
+          fetch-depth: 50
```
pzhlkj6612, line 20: Why did you modify "fetch-depth"?

.github/workflows/publish-to-pypi.yml
```diff
@@ -30,8 +30,8 @@ jobs:
         run: python -m build ./
       - name: twine check
         run: python -m twine check dist/*
-      - name: Publish distribution 📦 to Test PyPI
-        if: ${{ !startsWith(github.event.ref, 'refs/tags') }}
+      - name: Publish distribution 📦 to Test PyPI when releases branch
+        if: ${{ startsWith(github.event.ref, 'refs/tags/releases') }}
```
pzhlkj6612, line 34: I think you meant to write:

```suggestion
        if: ${{ startsWith(github.event.ref, 'refs/heads/releases') }}
```

There is no tag whose name starts with "releases", isn't it?

.github/workflows/publish-to-pypi.yml
miurahr: 2.0.0rc4 development went to .dev122, so I think it may be better to increase a depth for the test with test.pypi.org.
It is not mandatory.

miurahr: Oh, I also modify to run uploading to test.pypi.org on releases branch, that should not be such development count.
OK, we can keep it as is.

pzhlkj6612:
> * Check uploading  test.pypi.org only for releases branch

I checked https://github.com/miurahr/aqtinstall/branches and didn't find the branch named `releases/2.*` . Did you forget it when developing 2.0.0?

miurahr:
releases/1.2 branch was started when starting v2.0 development.
releases/2.0 branch will be started when v2.1 dev start, a day merging a feature targeted as 2.1 milestone.

pzhlkj6612:
> releases/1.2 branch was started when starting v2.0 development. releases/2.0 branch will be started when v2.1 dev start, a day merging a feature targeted as 2.1 milestone.

Got, thank you.

PR #414, opened 2021-09-30
Fix legacy commands

### LINKED ISSUE #413: Keeping v1 commands available in v2 may confuse users
## Overview

In [Legacy subcommands § Command Line Options — aqtinstall 2.0.0 documentation](https://aqtinstall.readthedocs.io/en/v2.0.0/cli.html#legacy-subcommands), we wrote:

> The subcommands `install`, `tool`, `src`, `doc`, and `examples` have been deprecated in favor of the newer `install-*` commands, but they remain in aqt in case you still need to use them.

After reading the above paragraph, I will think *"Fine, 'aqt tool' is still there. Those guys are keeping two series of commands in v2. I don't have to update my script after upgrading aqtinstall"*. However, the reality is that after switching to v2, the old script will **NOT** work without modification, at least for the "tool" command.

## Test

A working example for installing QIFW by aqtinstall 1.2.5:

```console
$ aqt tool linux tools_ifw 4.1.1 qt.tools.ifw.41
Specified target combination is not valid: linux tools_ifw qt.tools.ifw.41
...
Time elapsed: 8.37262754 second
```

The same script does not work in 2.0.0:

```console
$ aqt tool linux tools_ifw 4.1.1 qt.tools.ifw.41
usage: aqt tool ...
...
aqt tool: error: argument target: invalid choice: 'tools_ifw' (choose from 'desktop', 'winrt', 'android', 'ios')
Error: Process completed with exit code 2.
```

Rewrite the script according to the documentation of "[install-tool](https://aqtinstall.readthedocs.io/en/v2.0.0/cli.html#install-tool-command)" but keep the *legacy* "tool" command. It works in 2.0.0:

```diff
- aqt tool linux         tools_ifw 4.1.1 qt.tools.ifw.41
+ aqt tool linux desktop tools_ifw
```

```console
$ aqt tool linux desktop tools_ifw
aqtinstall(aqt) v2.0.0 on Python 3.8.10 [CPython GCC 9.4.0]
Warning: The command 'tool' is deprecated and marked for removal in a future version of aqt.
In the future, please use the command 'install-tool' instead.
...
Time elapsed: 11.00345962 second
```

Add the "tool variant name":

```diff
- aqt tool linux desktop tools_ifw
+ aqt tool linux desktop tools_ifw qt.tools.ifw.41
```

```console
$ aqt tool linux desktop tools_ifw qt.tools.ifw.41
aqtinstall(aqt) v2.0.0 on Python 3.8.10 [CPython GCC 9.4.0]
Warning: The command 'tool' is deprecated and marked for removal in a future version of aqt.
In the future, please use the command 'install-tool' instead.
...
Time elapsed: 9.82558412 second
```

*Use the command 'install-tool' instead*:

```diff
- aqt tool         linux desktop tools_ifw qt.tools.ifw.41
+ aqt install-tool linux desktop tools_ifw qt.tools.ifw.41
```

```console
$ aqt install-tool linux desktop tools_ifw qt.tools.ifw.41
aqtinstall(aqt) v2.0.0 on Python 3.8.10 [CPython GCC 9.4.0]
...
Time elapsed: 11.23552921 second
```

## Conclusion

As you can see, in aqtinstall v2, "tool" is not a legacy subcommand but **another name** of "install-tool". Those old interfaces have been changed, so keeping them available in v2 may confuse users.

I really know that the API changes can be (or should be) made in a new major release, but we didn't keep the availability and compatibility for those legacy commands, and what's worse is that we didn't document it correctly and clearly.

## Ignorable possible solutions

Here are some expensive ways to solve this problem:

- Remove those legacy commands in the next major version and release this new version as soon as possible. Yes, aqtinstall v3.
- Or, make those legacy commands usable again in the next minor version of v2. In other words, keep two series (v1 and v2) of commands in aqtinstall ^v2.1.

ddalcino:
Fix #413.

This fix will correctly map the legacy v1.* commands to the new v2.* commands, with backwards-compatible syntax.

Where necessary, the new commands should be modified to reproduce the legacy behavior when run in legacy mode. Right now, I think the only place we need to do this is `aqt tool`, where the legacy command would only install a tool if it matches the supplied version.

~~Right now, all I have are tests; implementation will follow.~~

**Edit**: forgot to implement legacy `aqt tool` behavior

**Edit**: implemented legacy `aqt tool` behavior that fails to install the tool when the version numbers don't match. Also re-implemented legacy bug that prevents installation of tools that don't have a valid semantic version. 💩 Hopefully, users who experience this bug will be motivated to use the new `aqt install-tool`.

We can roll back c5b89f6 if you want to keep the bug fixed, but this will remove the "fail on wrong version" behavior.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -3,6 +3,7 @@ import sys
 from pathlib import Path
 
 import pytest
+from pytest_socket import disable_socket
 
 from aqt.exceptions import CliInputError
 from aqt.installer import Cli
@@ -216,6 +217,51 @@ def test_cli_input_errors(capsys, expected_help, cmd, expect_msg, should_show_he
     out, err = capsys.readouterr()
     assert out == (expected_help if should_show_help else "")
     assert err.rstrip().endswith(expect_msg)
+
+
+# These commands use the new syntax with the legacy commands
+@pytest.mark.parametrize(
+    "cmd",
+    (
+        "install linux desktop 5.10.0",
+        "install linux desktop 5.10.0 gcc_64",
+        "src linux desktop 5.10.0",
+        "doc linux desktop 5.10.0",
+        "example linux desktop 5.10.0",
+        "tool windows desktop tools_ifw",
+        "tool windows desktop tools_ifw qt.tools.ifw.31",
+    ),
+)
+def test_cli_legacy_commands_with_wrong_syntax(cmd):
+    disable_socket()
+    cli = Cli()
+    cli._setup_settings()
+    with pytest.raises(SystemExit) as e:
+        cli.run(cmd.split())
+    assert e.type == SystemExit
+
+
+# These commands come directly from examples in the legacy documentation
+@pytest.mark.parametrize(
+    "cmd",
+    (
+        "install 5.10.0 linux desktop",  # default arch
+        "install 5.10.2 linux android android_armv7",
+        "src 5.15.2 windows desktop --archives qtbase --kde",
+        "doc 5.15.2 windows desktop -m qtcharts qtnetworkauth",
+        "examples 5.15.2 windows desktop -m qtcharts qtnetworkauth",
+        "tool linux tools_ifw 4.0 qt.tools.ifw.40",
+    ),
+)
+def test_cli_legacy_commands_with_correct_syntax(monkeypatch, cmd):
+    # Pretend to install correctly when any command is run
+    for func in ("run_install_qt", "run_install_src", "run_install_doc", "run_install_example", "run_install_tool"):
+        monkeypatch.setattr(Cli, func, lambda *args, **kwargs: 0)
+
+    disable_socket()
+    cli = Cli()
+    cli._setup_settings()
+    assert 0 == cli.run(cmd.split())
 
 
 def test_cli_unexpected_error(monkeypatch, capsys):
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -218,6 +219,51 @@ def test_cli_input_errors(capsys, expected_help, cmd, expect_msg, should_show_he
...
+@pytest.mark.parametrize(
+    "cmd",
+    (
+        "install linux desktop 5.10.0",
+        "install linux desktop 5.10.0 gcc_64",
+        "src linux desktop 5.10.0",
+        "doc linux desktop 5.10.0",
+        "example linux desktop 5.10.0",
+        "tool windows desktop tools_ifw",
+        "tool windows desktop tools_ifw qt.tools.ifw.31",
```
ddalcino, line 232: This is the wrong way to test `aqt tool`; I'm not sure how to fix it yet. This test relies on `argparse` to raise `SystemExit` when the arguments are invalid. `aqt tool` is supposed to take 4 positional arguments and `argparse` does not validate the last 3. 

If you run `aqt tool windows desktop tools_ifw qt.tools.ifw.31`, the correct behavior is for `aqt tool` to try to fetch a tool called `desktop` with version `tools_ifw` and variant `qt.tools.ifw.31`.

miurahr:
Thank you!!!!

PR #415, opened 2021-10-01
Fix usage of `disable_socket`

ddalcino:
In an earlier PR, I added calls to `disable_socket()` from `pytest_socket` where I thought they were needed to prevent some tests from accessing the network, in case they weren't monkeypatched properly. Today, I discovered that `disable_socket()` disables sockets globally for all tests, which means that the tests that use remote data cannot run if they are executed after another test calls `disable_socket()`.

This change calls `disable_socket()` once from `conftest.py`, so that no tests are allowed to use network data unless they are marked as OK to use the network, with `@pytest.mark.enable_socket`. See example of usage in `tests/test_connection.py`.

See [pytest_socket docs](https://github.com/miketheman/pytest-socket).

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -64,6 +64,8 @@ class Cli:
 
     __slot__ = ["parser", "combinations", "logger"]
 
+    UNHANDLED_EXCEPTION_CODE = 254
+
     def __init__(self):
         parser = argparse.ArgumentParser(
             prog="aqt",
@@ -112,7 +114,7 @@ class Cli:
                 "Please file a bug report at https://github.com/miurahr/aqtinstall/issues.\n"
                 "Please remember to include a copy of this program's output in your report."
             )
-            return 1
+            return Cli.UNHANDLED_EXCEPTION_CODE
 
     def _check_tools_arg_combination(self, os_name, tool_name, arch):
         for c in Settings.tools_combinations:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,5 @@
+from pytest_socket import disable_socket
+
+
+def pytest_runtest_setup():
+    disable_socket()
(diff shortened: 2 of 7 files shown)


ddalcino:
Test failed due to checksum error while downloading an archive. This is unrelated to the changed code.

aqt/installer.py
```diff
@@ -100,6 +100,10 @@ def run(self, arg=None) -> int:
                 self.show_help()
             return 1
         except Exception as e:
+            # For testing: Allow blocked socket errors to cause failed tests
+            if e.__class__.__name__ in ("SocketBlockedError", "SocketConnectBlockedError"):
+                raise e from e
+
```
ddalcino, line 106: I think this was a really bad idea, and I'm going to change it.

The problem that this is trying to solve is: During tests, how do we differentiate between an exception that was caught properly (via the `except AqtException` clause above) vs. an exception that was unexpected? If I write a test that checks the return code (1 or 0), and I expect the return code to be 1, I have no idea which exception handler caught the exception unless I read from stderr. What I really need is an easy way to check whether or not I triggered the `except Exception` branch, without reading stderr.

The straightforward solution would be something like this:
```python
try:
  args.func(args)
  # success
  return 0
except AqtException:
  # expected error
  return 1
except Exception:
  # catastrophic unexpected error
  return 2  # or some other number
```
This way, we can write write tests that can use the return value to check for unexpected events, rather than reading stderr.

The only problem with this is that I'm not sure what error code to use for "catastrophic unexpected error". I don't need this number to be 2, but I need it to be something other than 0 or 1.

miurahr: For bash script, 2 means misuse of shell-builtin.
https://tldp.org/LDP/abs/html/exitcodes.html

There seems to be a range which has a special meaning, 126-165, 255.
It may be better to use a number other than these.

ddalcino: How do you feel about 254?

miurahr: 254 looks OK

miurahr:
LGTM.
I'd like to squash and merge it.

Issue #416, opened 2021-10-01
MemoryError during decompression & Unable to read c:\Users\qt\work\install\mkspecs\qconfig.pri


dwheeld:
A recent install gave this bug report:

  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 948, in extractall
    self._extract(path=path, return_dict=False, callback=callback)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 604, in _extract
    self.worker.extract(
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 1165, in extract
    self.extract_single(
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 1276, in extract_single
    raise e
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 1261, in extract_single
    crc32 = self.decompress(fp, f.folder, ofp, f.uncompressed, f.compressed, src_end)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/py7zr.py", line 1313, in decompress
    tmp = decompressor.decompress(fp, out_remaining)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/compressor.py", line 630, in decompress
    tmp = self._decompress(self._unused + data, max_length)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/py7zr/compressor.py", line 602, in _decompress
    data = decompressor.decompress(data)
MemoryError
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:/msys64/mingw64/lib/python3.9/site-packages/aqt/installer.py", line 99, in run
    args.func(args)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/aqt/installer.py", line 294, in run_install_qt
    run_installer(qt_archives.get_packages(), base_dir, sevenzip, keep)
  File "C:/msys64/mingw64/lib/python3.9/site-packages/aqt/installer.py", line 835, in run_installer
    pool.starmap(installer, tasks)
  File "C:/msys64/mingw64/lib/python3.9/multiprocessing/pool.py", line 372, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "C:/msys64/mingw64/lib/python3.9/multiprocessing/pool.py", line 771, in get
[...]

PR #419, opened 2021-10-03
Propagate exceptions from `run_installer()`

ddalcino:
This change catches exceptions derived from Exception and KeyboardInterrupt raised by `aqt.installer.installer()`, while run by multiple processes, and propagates them back to earlier stack entries. This will prevent any OSError and BrokenPipe exceptions that would otherwise be raised when one process has an exception while the other processes are still running.

This also handles the MemoryError exception we saw in #416, and offers some suggestions for solving the issue.


aqt/installer.py
```diff
@@ -824,21 +824,38 @@ def run_installer(archives: List[QtPackage], base_dir: str, sevenzip: Optional[s
         tasks.append((arc, base_dir, sevenzip, queue, keep))
     ctx = multiprocessing.get_context("spawn")
     pool = ctx.Pool(Settings.concurrency, init_worker_sh)
+
+    def close_worker_pool_on_exception(exception: BaseException):
+        logger = getLogger("aqt.installer")
+        logger.warning(f"Caught {exception.__class__.__name__}, terminating installer workers")
+        pool.terminate()
+        pool.join()
```
ddalcino, line 832: It may be possible to move this block into the `finally` clause, so we don't have to call it in each `except` clause. I'm not sure it's worth doing though.

miurahr: No need to move. Current is better to maintain.

PR #420, opened 2021-10-03
Allow retry of download on checksum error

ddalcino:
This allows downloads of archives to be retried when checksum errors occur, as discussed in https://github.com/miurahr/aqtinstall/issues/416#issuecomment-932472110. Even after I [turned off my VPN](https://github.com/miurahr/aqtinstall/issues/416#issuecomment-933031438), I am still getting so many checksum errors that it takes me several tries before I can successfully install Qt. If any of our users are having similar trouble, I'm sure that they are extremely frustrated with `aqt`, and I think that this change could be helpful for them.

I'm not certain that this is the best implementation of this feature; please see my notes below.

### DIFF
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -77,7 +77,7 @@ def getUrl(url: str, timeout) -> str:
     return result
 
 
-def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
+def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: bytes, timeout):
     logger = getLogger("aqt.helper")
     filename = Path(url).name
     with requests.Session() as session:
@@ -109,7 +109,8 @@ def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
             if exp is not None and hash.digest() != exp:
                 raise ArchiveChecksumError(
                     f"Downloaded file {filename} is corrupted! Detect checksum error.\n"
-                    f"Expected {exp}, Actual {hash.digest()}"
+                    f"Expect {exp.hex()}: {url}\n"
+                    f"Actual {hash.digest().hex()}: {out}"
                 )
 
 
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -28,13 +28,14 @@ import sys
 import xml.etree.ElementTree as ElementTree
 from logging import getLogger
 from logging.handlers import QueueListener
-from typing import Callable, Dict, List
+from pathlib import Path
+from typing import Callable, Dict, List, Tuple
 from urllib.parse import urlparse
 
 import requests
 import requests.adapters
 
-from aqt.exceptions import ArchiveConnectionError, ArchiveDownloadError, ArchiveListError
+from aqt.exceptions import ArchiveChecksumError, ArchiveConnectionError, ArchiveDownloadError, ArchiveListError
 
 
 def _get_meta(url: str):
@@ -78,6 +79,7 @@ def getUrl(url: str, timeout) -> str:
 
 def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
     logger = getLogger("aqt.helper")
+    filename = Path(url).name
     with requests.Session() as session:
         retries = requests.adapters.Retry(total=Settings.max_retries, backoff_factor=Settings.backoff_factor)
         adapter = requests.adapters.HTTPAdapter(max_retries=retries)
@@ -103,11 +105,28 @@ def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
                         hash.update(chunk)
                     fd.flush()
             except Exception as e:
-                raise ArchiveDownloadError(f"Download error: {e}") from e
+                raise ArchiveDownloadError(f"Download of {filename} has error: {e}") from e
             if exp is not None and hash.digest() != exp:
-                raise ArchiveDownloadError(
-                    f"Download file is corrupted! Detect checksum error.\nExpected {exp}, Actual {hash.digest()}"
+                raise ArchiveChecksumError(
+                    f"Downloaded file {filename} is corrupted! Detect checksum error.\n"
+                    f"Expected {exp}, Actual {hash.digest()}"
                 )
+
+
+def retry_on_errors(action: Callable[[], any], acceptable_errors: Tuple, num_retries: int, name: str):
+    logger = getLogger("aqt.helper")
+    for i in range(num_retries):
+        try:
+            retry_msg = f": attempt #{1 + i}" if i > 0 else ""
+            logger.info(f"{name}{retry_msg}...")
+            ret = action()
+            if i > 0:
+                logger.info(f"Success on attempt #{1 + i}: {name}")
+            return ret
+        except acceptable_errors as e:
+            if i < num_retries - 1:
+                continue  # just try again
+            raise e from e
 
 
 def altlink(url: str, alt: str):
@@ -314,6 +333,10 @@ class SettingsClass:
     @property
     def max_retries(self):
         return self.config.getfloat("requests", "max_retries", fallback=5)
+
+    @property
+    def max_retries_on_checksum_error(self):
+        return self.config.getint("requests", "max_retries_on_checksum_error", fallback=int(self.max_retries))
 
     @property
     def backoff_factor(self):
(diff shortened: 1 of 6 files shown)


aqt/helper.py
```diff
@@ -103,13 +105,30 @@ def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
...
                     fd.flush()
             except Exception as e:
-                raise ArchiveDownloadError(f"Download error: {e}") from e
+                raise ArchiveDownloadError(f"Download of {filename} has error: {e}") from e
             if exp is not None and hash.digest() != exp:
-                raise ArchiveDownloadError(
-                    f"Download file is corrupted! Detect checksum error.\nExpected {exp}, Actual {hash.digest()}"
+                raise ArchiveChecksumError(
+                    f"Downloaded file {filename} is corrupted! Detect checksum error.\n"
+                    f"Expected {exp}, Actual {hash.digest()}"
```
ddalcino, line 112: I found that it was useful to report the filename in the error message. This isn't necessary, but every time I got the checksum error, I wanted to know which file caused it.

aqt/helper.py
```diff
@@ -103,13 +105,30 @@ def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: str, timeout):
...
+            retry_msg = f": attempt #{1 + i}" if i > 0 else ""
+            logger.info(f"{name}{retry_msg}...")
+            ret = action()
+            if i > 0:
+                logger.info(f"Success on attempt #{1 + i}: {name}")
+            return ret
+        except acceptable_errors as e:
+            if i < num_retries - 1:
+                continue  # just try again
+            raise e from e
```
ddalcino, line 129: This is the simplest way I could think of to write the "retry" function, while still logging some output when it's useful. I like keeping it separate from `downloadBinaryFile`, for readability and reusability, but we could move this logic directly into that function if you think that would be better.

The `acceptable_errors` parameter is a little weird, and very easy to misuse. It must contain a tuple of Exception classes, or the function will fail. I thought it was useful to pass the exceptions in this way; that's the only way this function can be reused effectively. On the other hand, this is such a strange thing to do, I'm concerned that this could be an antipattern.

aqt/helper.py
```diff
@@ -315,6 +334,10 @@ def response_timeout(self):
     def max_retries(self):
         return self.config.getfloat("requests", "max_retries", fallback=5)
 
+    @property
+    def max_retries_on_checksum_error(self):
+        return self.config.getint("requests", "max_retries_on_checksum_error", fallback=int(self.max_retries))
```
ddalcino, line 339: This adds an extra setting to `settings.ini` that allows you to configure how many times `aqt` will retry a download when a checksum error occurs. I think that it's useful to differentiate between this and `max_retries`, which determines how many times a connection should be retried; these are different concepts and should be handled differently.

I worry that it will be confusing to a user to have these two settings; the difference is unclear unless you read the code. Maybe `max retries` could be renamed.

I don't know if falling back to `max_retries` is wise, or even makes sense.

aqt/helper.py
pzhlkj6612: Is this filename a filename, a relative path or an absolute path?

ddalcino: Just the filename. `Path(url).name` is just a really quick way to extract the filename, without the preceding path. We could print the whole url, but all I really wanted to see is the filename.

aqt/helper.py
miurahr: We can mark `max_retries` as deprecated and introduce new parameter `max_retries_on_connection_error` that fallback to `max_retries`. Does it help users?

aqt/helper.py
pzhlkj6612: > Just the filename. `Path(url).name` is just a really quick way to extract the filename, without the preceding path.

OK, I didn't read the code before asking that dumb question.

> We could print the whole url, but all I really wanted to see is the filename.

I found that the whole url has already been printed by [`logger.debug("Download URL: {}".format(url))` at installer.py#L899](https://github.com/miurahr/aqtinstall/blob/4eed29fdf1a8c38af73e3fa547eac8fdea327271/aqt/installer.py#L899). However, aqt downloads files in parallel, so the output may be scattered (will it?). In my opinion, it may be more helpful to print the whole URL again, and print the destination path `out` as well. If so, due to the complete URL and path, you would be able to find the local file and manually re-download it from the server more quickly and easily.

The output could be this:

```
...
Downloaded file ${filename} is corrupted! Detect checksum error.
${exp}:           ${url}
${hash.digest()}: ${out}
```

Note that I want to align the url and the path. The length of hash strings are fixed so it can be achieved.

aqt/helper.py
ddalcino: > We can mark `max_retries` as deprecated and introduce new parameter `max_retries_on_connection_error` that fallback to `max_retries`. Does it help users?

That sounds like a good idea. So here's how `Settings` would work:
1. Try to load `max_retries`, and fallback to 5.
2. Try to load `max_retries_on_connection_error`, and fallback to `max_retries`.
3. Try to load `max_retries_on_checksum_error`, and fallback to `max_retries`.

I'm not sure how to mark `max_retries` as deprecated; would we do that in the documentation somewhere?

aqt/helper.py
pzhlkj6612: > I'm not sure how practical it is to manually re-download the file; when this error occurs, there could be several other files that have not even started downloading yet.

I mean, if we print the URL, developers or experienced users will be able to directly copy-paste it to their browser or HTTP downloader to manually download the file, then do whatever they like.

> How about this for the error message:
> 
> ```
> ...
> Downloaded file ${filename} is corrupted! Detect checksum error.
> Expect ${exp.hex()}:           ${url}
> Actual ${hash.digest().hex()}: ${out}
> ```
> 
> I like having the "Expect" and "Actual" labels there just to clarify which is which.

I agree.

> When `installer()` calls this, it just passes the name of the archive file into `out` without a path, so the file downloads to the working directory. Since `out` has no path info attached to it, I can't tell what it's supposed to be.

OK, I understand it. Those 7z archives contain the proper structure of files and directories, so we can download them to the current working directory (any writable locations are OK) and then extract the content to the specific destination (CWD or `outputdir` ).

How about `str(Path.cwd() / out)` ?

> Both `exp` and `hash.digest()` are "bytes" objects, and if you print them as-is, they look like "\x01\x\23\x45\x67...". The `.hex()` would turn the same value into "01234567". It's half as long, and I think it's easier to read.

Yes, it should definitely be a string. I forgot what the `hash.digest()` function returns.

ddalcino: > How about `str(Path.cwd() / out)`?

Yes, that would be an improvement.

I am a little unclear on why we are downloading to the working directory in the first place. If the user happens to have files in the working directory with the same names as the archives downloaded, these files will be overwritten and then deleted. Additionally, if `aqt` fails before finishing this process, the working directory will be littered with `.7z` files that may or may not be partially downloaded. 

I think it might be more respectful of the user's disk space to use a temporary directory to download these files. A temp directory would delete itself automatically when the program ends, whether an error occurs of not. However, this would make it more difficult to attempt to recover an unfinished installation.

miurahr:
Thank you for all the improvements! Now it looks ready to merge.

aqt/helper.py
pzhlkj6612: > ...
> 
> I think it might be more respectful of the user's disk space to use a temporary directory to download these files. A temp directory would delete itself automatically when the program ends, whether an error occurs of not.

I agree. I think downloading archives to `%TMP%` by default makes sense.

This is worth filing a new issue.

> However, this would make it more difficult to attempt to recover an unfinished installation.

Add a new option (parameter) to allow the user to specify the download location? Combine with the `-k, --keep` option (parameter)?

Also, new issue please.

pzhlkj6612: > > ...
> > I think it might be more respectful of the user's disk space to use a temporary directory to download these files. A temp directory would delete itself automatically when the program ends, whether an error occurs of not.
> 
> I agree. I think downloading archives to `%TMP%` by default makes sense.
> 
> This is worth filing a new issue.

Done. See #421 .

> > However, this would make it more difficult to attempt to recover an unfinished installation.
> 
> Add a new option (parameter) to allow the user to specify the download location? Combine with the `-k, --keep` option (parameter)?
> 
> Also, new issue please.

Discussion #422 . Talking in the discussion seems more suitable.

PR #446, opened 2021-11-05
fix failing coverage upload

ddalcino:
Recently, every "Upload Coverage" step fails for me. This should fix it.

Example of failed "Upload Coverage" step: https://github.com/ddalcino/aqtinstall/runs/4122878621?check_suite_focus=true#step:8:31


miurahr:
Thank you!

Issue #436, opened 2021-10-29
Memory error is occurred when running aqt install-qt windows desktop 6.2.0 on 32bit python on windows


martinrotter:
```
 λ aqt list-qt windows desktop --modules 6.2.0 win64_msvc2019_64
debug_info qt3d qt5compat qtactiveqt qtcharts qtconnectivity qtdatavis3d qtimageformats qtlottie qtmultimedia qtnetworkauth qtpositioning qt

quick3d qtquicktimeline qtremoteobjects qtscxml qtsensors qtserialbus qtserialport qtshadertools qtvirtualkeyboard qtwebchannel qtwebengine

qtwebsockets qtwebview

 /cygdrive/c/Users/rotter/Downloads/PraktikSW.Praktik.cz  chyba ❗
 λ aqt install-qt --outputdir C:\\Qt windows desktop 6.2.0 win64_msvc2019_64 -m qtwebengine qtmultimedia qtwebchannel qt5compat qtimageformats
aqtinstall(aqt) v2.0.0 on Python 3.8.3 [CPython MSC v.1925 32 bit (Intel)]
Downloading qtimageformats...
Downloading qtmultimedia...
Downloading qtwebchannel...
Downloading qtwebengine...
Redirected: ftp.fau.de
Redirected: ftp.fau.de
Redirected: ftp.fau.de
Redirected: ftp.fau.de
Finished installation of qtwebchannel-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 0.60492760
Downloading qt5compat...
Finished installation of qtimageformats-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 0.89102150
Downloading qtbase...
Finished installation of qtmultimedia-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 0.97993880
Downloading qtsvg...
Redirected: ftp.fau.de
Redirected: ftp.fau.de
Redirected: ftp.fau.de
Finished installation of qt5compat-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 0.86194740
Downloading qtdeclarative...
Finished installation of qtsvg-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 0.71559730
Downloading qttools...
Redirected: ftp.fau.de
Redirected: ftp.fau.de
[...]

PR #447, opened 2021-11-08
CI: add test with 32bit python on windows

miurahr:
Work for #436


miurahr:
This does not reproduce the case #436

PR #450, opened 2021-11-14
Fix cross-platform installation failure

### LINKED ISSUE #449: Unable to download windows mingw binaries on Linux host
Hello,

I am trying to setup a Docker container to cross-compile Windows binaries on a Linux host using the MinGW toolchain. Here is my Dockerfile so far :

```dockerfile
FROM --platform=linux/amd64 debian:buster

RUN apt-get update

RUN apt-get install -y python3-pip

RUN pip3 install aqtinstall

RUN aqt install-qt windows desktop 5.15.2 win64_mingw81 --outputdir /opt/qt --archives qtbase
```

The aqt step fails with the following error : 

```
aqtinstall(aqt) v2.0.2 on Python 3.7.3 [CPython GCC 8.3.0]
Downloading qtbase...
Redirected: ftp2.nluug.nl
Finished installation of qtbase-Windows-Windows_10-Mingw-Windows-Windows_10-X86_64.7z in 44.70309867
Updater caused an IO error: [Errno 13] Permission denied: '/opt/qt/5.15.2/mingw81_64/bin/qmake.exe'
The command '/bin/sh -c aqt install-qt windows desktop 5.15.2 win64_mingw81 --outputdir /opt/qt --archives qtbase' returned a non-zero code: 1
```

It looks like aqt is trying to execute `/opt/qt/5.15.2/mingw81_64/bin/qmake.exe`, which indeed doesn't have execute permission, and which can't execute anyway since we're on a Linux host.

Am I doing something wrong ? Why is aqt trying to execute qmake ?

Thanks.

ddalcino:
Catch IOError or OSError that raised by launching `qmake -query` when cross platform installation.

Fix #449

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -80,7 +80,7 @@ class Updater:
                 continue
             try:
                 result = subprocess.run([str(qmake_path), "-query"], stdout=subprocess.PIPE)
-            except subprocess.SubprocessError:
+            except (subprocess.SubprocessError, IOError, OSError):
                 return False
             if result.returncode == 0:
                 self.qmake_path = qmake_path


ddalcino:
Oops! The `install-tool` test failed because QtIFW 4.1 was replaced with 4.2. I forgot to fix this earlier, so I guess now would be a good time. #451 should fix it.

Issue #453, opened 2021-11-15
Bug in install-src


cradcore:
I am trying to install the source files for QtWebEngine, linux desktop 5.9.9 gcc_64. However, this is what I get:

```
cradcore@localhost: ~ $ aqt install-src -m qtwebengine -O /dev-tools linux desktop 5.9.9
aqtinstall(aqt) v2.0.2 on Python 3.8.10 [CPython GCC 9.3.0]
The packages ['qtwebengine'] were not found while parsing XML of package information!
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt linux desktop --modules 5.9.9 <arch>' to show modules available.
cradcore@localhost: ~ $ aqt list-qt linux desktop --modules 5.9.9 gcc_64
qtcharts qtdatavis3d qtnetworkauth qtpurchasing qtremoteobjects qtscript qtspeech qtvirtualkeyboard qtwebengine
```

It seems to not find qtwebengine when trying to install, but correctly lists it as a module

ddalcino:
Looks like the "Suggested follow-up" message in this case is inappropriate. The `aqt list-qt` command correctly shows output pertinent to the `install-qt` command; it is not designed to provide data relevant to `install-src/doc/example`.

See the Updates.xml file at https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/qt5_599_src_doc_examples/Updates.xml: there is no source module for qtwebengine. The archive `qtwebengine-opensource-src-5.9.9.7z` is part of the `qt.qt5.599.src` PackageUpdate; that means that you can install it by running `aqt install-src linux desktop 5.9.9`.

As far as I can tell, there are no valid options that you can pass to the `-m` flag for `install-src`. Unless I have overlooked something, maybe we should remove the flag.

The `list-qt` command is technically capable of to showing data relevant to `install-src/doc/example`, by using `--extension src_doc_examples`, but it doesn't list modules and it isn't documented properly. Right now, these are the only commands that work:
```
aqt list-qt linux desktop --extension src_doc_examples --archives 5.9.9 src
aqt list-qt linux desktop --extension src_doc_examples --archives 5.9.9 doc
aqt list-qt linux desktop --extension src_doc_examples --archives 5.9.9 examples
```

### Todo:
1. Remove/replace the "Suggested follow-up" message for `install-src/doc/example` when a module cannot be located
2. Consider adding commands for `list-src/doc/example`.
[...]

PR #454, opened 2021-11-23
Add `list-src/doc/example` command

ddalcino:
The `list-src`, `list-doc`, and `list-example` commands will print all the modules and archives available for use with the `install/src/doc/example` commands.

The impetus for this PR comes from #453.


ddalcino:
I tried to split my fix for #453 into two discrete parts; looks like the split failed. I will submit a more complete PR instead; if you feel like it should be split in half, I will fix this PR and reopen it.

PR #455, opened 2021-11-23
Fix problems with `install-src/doc/examples`

### LINKED ISSUE #453: Bug in install-src

ddalcino:
This PR is doing a lot of work, and I understand if you would like me to split it into smaller parts:

1. Add the commands `list-src`, `list-doc`, and `list-example`, to print modules and archives for source, docs, and examples
2. Deprecate the `target` parameter for `install-src/doc/example`
3. Remove the `--modules` parameter from `install-src`
4. Update the error messages that ask the user to use `list-qt` to show valid options, so that they print the correct `list-src/doc/example` command. ~~(not implemented yet)~~ implemented in cfc2410

Fix #453


ddalcino:
I am confused about what happened in this build: https://dev.azure.com/miurahr/github/_build/results?buildId=4747&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=777e4b2c-7cad-5027-3b62-2efbb20eb961

This is a failure that I have never seen before. It occurs during installation of Pscx, in this code:
https://github.com/miurahr/aqtinstall/blob/60018b3bf643b8bd2901bd2102c9bf1e53ed888b/ci/steps.yml#L163-L167

[This build](https://dev.azure.com/miurahr/github/_build/results?buildId=4746&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=777e4b2c-7cad-5027-3b62-2efbb20eb961) passed when it ran in the previous commit, and the code in this commit is completely unrelated to the build step that failed.

I would speculate that this failed due to a backpressure mechanism at the NuGet server. I would like to re-run the build and see if it is still broken, but I'm not sure how to do that.

ci/steps.yml
```diff
@@ -108,12 +108,28 @@ steps:
         fi
         aqt list-qt $(HOST) $(TARGET) $ext --arch $(QT_VERSION)                 # print architectures for version/host/target
         aqt list-qt $(HOST) $(TARGET) $ext --spec "$(SPEC)" --arch latest
+
+        # list-src/doc/example
+        aqt list-src $(HOST) $(QT_VERSION)                                      # print source archives available for host/desktop/version
+        aqt list-doc $(HOST) $(QT_VERSION)                                      # print docs archives available for host/desktop/version
+        aqt list-doc $(HOST) $(QT_VERSION) --modules                            # print docs modules available for host/desktop/version
+        aqt list-example $(HOST) $(QT_VERSION)                                  # print example archives available for host/desktop/version
+        aqt list-example $(HOST) $(QT_VERSION) --modules                        # print example modules available for host/desktop/version
```
ddalcino, line 117: This code demonstrates how the new `list-src`, `list-doc`, and `list-example` commands work. I am open to suggestions on how they should be designed; I am not certain that this was the right choice.

There are two obvious options here:
1. list modules by default, and use `--archives` to print archives
2. list archives by default, and use `--modules` to print modules

I used option 2 because I wanted the interface to be consistent across `list-src`, `list-doc`, and `list-example`. The `list-src` command cannot print modules at all, so if I had used option 1, `list-src mac 6` would print an error message, and `list-src mac 6 --archives` would print archives; I thought this would be surprising.

ci/generate_azure_pipelines_matrices.py
```diff
@@ -201,10 +201,15 @@ def __init__(self, platform, build_jobs):
...
+            "install-src", "6.1.0", "linux", "desktop", "gcc_64", "gcc_64", subarchives="qtlottie"
         ),
         BuildJob(
-            "install-doc", "6.1.0", "linux", "desktop", "gcc_64", "gcc_64", subarchives="qtdoc"
+            "install-doc", "6.1.0", "linux", "desktop", "gcc_64", "gcc_64",
+            subarchives="qtdoc qtlottieanimation", module="qtlottie"
+        ),
+        BuildJob(
+            "install-example", "6.1.0", "linux", "desktop", "gcc_64", "gcc_64",
+            subarchives="qtdoc qtlottie", module="qtlottie"
```
ddalcino, line 212: This code illustrates a potential problem with `list-doc`/`list-example`: These commands cannot list archives for modules. This is a problem for the `qtlottie` module, because the archive is called `qtlottieanimation` when you run `install-doc`, but it's called `qtlottie` everywhere else.

We could add a function that prints the names of archives for each module, the way we did for `list-qt`. I don't like this idea, because I think it would be better to just make sure that modules are installed by default whenever you ask for them. I'll open an issue to track this; I prefer to solve this problem in a separate PR.

ddalcino: **Please note:** The module `qtlottie` is a bad choice for `install-example` with Qt 6.1.0; the archive is empty!

```console
$ 7z l 6.1.0-0-202105040937qtlottie-examples-6.1.0.7z

7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=C.UTF-8,Utf16=on,HugeFiles=on,64 bits,6 CPUs Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz (906EA),ASM,AES-NI)

Scanning the drive for archives:
1 file, 162 bytes (1 KiB)

Listing archive: /mnt/d/Downloads/6.1.0-0-202105040937qtlottie-examples-6.1.0.7z

--
Path = /mnt/d/Downloads/6.1.0-0-202105040937qtlottie-examples-6.1.0.7z
Type = 7z
Physical Size = 162
Headers Size = 162
Solid = -
Blocks = 0

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2021-05-04 01:37:31 D....            0            0  Examples
2021-05-04 01:43:19 D....            0            0  Examples/Qt-6.1.0
------------------- ----- ------------ ------------  ------------------------
2021-05-04 01:43:19                  0            0  0 files, 2 folders
```
I chose this module because of the archive's small file size; I didn't realize there wasn't anything in it!

I should have seen this in Updates.xml:
```xml
<PackageUpdate>
  <Name>qt.qt6.610.examples.qtlottie</Name>
...
  <UpdateFile OS="Any" CompressedSize="202" UncompressedSize="40"/>
...
</PackageUpdate>
```

Maybe we could print a warning when `PackageUpdate.UpdateFile.CompressedSize` is greater than `PackageUpdate.UpdateFile.UncompressedSize`, or when UncompressedSize is less than some threshold value? It could be something like "This module is only 40 bytes when uncompressed; it might be empty!"

ddalcino: `qt.qt6.610.examples.qtimageformats` has this issue as well.

ddalcino: Resolved in d6a6f79: this prevents the `list-*` commands from printing any module <= 40 bytes when decompressed. The 40 bytes value can be adjusted in `settings.ini`. I don't think this will be a popular thing to change, so I will not make a CLI flag for it.

I did not change the behavior of `install-*` for this. You can still install empty modules, but you won't see them with `list-*` unless you change your settings.

This will cause `combinations.json` to omit empty modules, which means that `install-*` will warn the user that the empty modules are unrecognized.

miurahr:
LTGM.

ci/generate_azure_pipelines_matrices.py
miurahr: > any module <= 40 bytes when decompressed. 

It's nice findings how to detect empty modules.

PR #458, opened 2021-11-24
Feature: allow user to specify the download location of 7z archives

### LINKED ISSUE #421: Feature: specify the download location of 7z archives
I am a little unclear on why we are downloading to the working directory in the first place. If the user happens to have files in the working directory with the same names as the archives downloaded, these files will be overwritten and then deleted. Additionally, if `aqt` fails before finishing this process, the working directory will be littered with `.7z` files that may or may not be partially downloaded. 

I think it might be more respectful of the user's disk space to use a temporary directory to download these files. A temp directory would delete itself automatically when the program ends, whether an error occurs of not. However, this would make it more difficult to attempt to recover an unfinished installation.

Originally posted by @ddalcino in https://github.com/miurahr/aqtinstall/pull/420#discussion_r721494482

ddalcino:
Adds the option `--archive-dest <path>`.

For convenience/readability, this will also turn any reference to the downloaded file in `helper.py` and `installer.py` into a `pathlib.Path` object. This makes it easier to ensure that the file ends up in the right location.

This also adds items in `settings.ini` that control the default behavior of `--keep` and `--archive-dest`.

Fix #421

### DIFF
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -79,7 +79,7 @@ def getUrl(url: str, timeout) -> str:
     return result
 
 
-def downloadBinaryFile(url: str, out: str, hash_algo: str, exp: bytes, timeout):
+def downloadBinaryFile(url: str, out: Path, hash_algo: str, exp: bytes, timeout):
     logger = getLogger("aqt.helper")
     filename = Path(url).name
     with requests.Session() as session:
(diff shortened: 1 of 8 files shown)


miurahr:
LGTM. When test passed, I'd like to merge.

PR #459, opened 2021-11-25
Restrict effect of `--archives` flag to base and debug_info modules

### LINKED ISSUE #457: Change behavior of `--archives` so that it only affects the base module
Several modules for Qt, Docs, and Examples include archives that do not have the same name as the module. For example:
* Qt 6.2.1 for linux desktop includes the module `qtwaylandcompositor`, which consists of the archive `qtwayland`.
* Documentation for Qt 6.1.0 includes the module `qtlottie`, which consists of the archive `qtlottieanimation`. The archive is called `qtlottie` in most other places.

We already have some code that attempts to make sure that if the user has used both the `--archives` and `--modules` flags, the user should receive the archives that come from the requested archives:
https://github.com/miurahr/aqtinstall/blob/d63bbd146c6852b81caca72ab6abe9aa8c2962f2/aqt/installer.py#L269-L270

This approach will not install modules that include archives with a different name (like the examples listed above), and it won't generate an error message. I think that any user that experiences this situation will see it as a bug, rather than a misuse of `aqt`.

I would like to change the design of this feature, so that the `--archives` flag will only apply to the base Qt module (ie, the PackageUpdate called `qt.qt6.621.clang_64` in https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt6_621/Updates.xml), and to the `debug_info` module. We have already changed the design of `--noarchives` so that it only turns off the base Qt module; I think it makes sense to make `--archives` work the same way.

I think it should be simple to implement this in `aqt.archives.QtArchives._parse_update_xml()`. The QtArchives object recieves the list of desired archives as a parameter to `__init__`; it just needs to be told which modules include archives that need to be filtered.

ddalcino:
This changes the behavior of `--archives` so that it will affect the base module and the debug_info module, but no other modules.

Without this change, the following commands would fail to install the requested modules, and would not generate error messages. This change would allow these command to install the requested modules without a problem.

```bash
aqt install-qt linux desktop 5.15.2 --modules qtcharts --archives qtbase           # did not specify archive `qtcharts`
aqt install-doc linux desktop 6.1.0 --modules qtlottie --archives qtdoc qtlottie   # wrong archive name for qtlottie
                                                                   # the correct archive name is `qtlottieanimation`
```

Fix #457

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -436,6 +436,7 @@ of archives that it is downloading, extracting, and installing,
 including ``qtbase``, ``qtmultimedia``, ``qt3d``, and ~25 more items.
 We can use the ``--archives`` flag to choose which of these archives we will
 actually install.
+The ``--archives`` flag can only affect two modules: the base Qt installation and the ``debug_info`` module.
 
 .. note::
     In this documentation, **"modules"**, **"archives"**, and **"the base Qt installation"**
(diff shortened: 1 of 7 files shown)


ddalcino:
```
 _______________________________
< Happy Thanksgiving Everybody! >
 -------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

docs/getting_started.rst
```diff
@@ -511,72 +512,38 @@ working Qt installation. We can install these archives using this command:
 Installing Modules With Archives Specified
 ``````````````````````````````````````````
 
-Now let's say we need to install the ``qtcharts`` and ``qtlottie`` modules.
-Let's see what archives are part of these modules:
+As of aqt v2.1.0, the ``--archives`` flag will only apply to
```
ddalcino, line 515: I’m just guessing at what version this would be part of; please let me know if I should change “v2.1” to something else.

ddalcino:
Are we not going to use this PR? I thought it was flagged for v2.1

miurahr:
I've waited all the check passed. I'd like to run test again.

PR #464, opened 2021-12-06
Reduce memory consumption on 32bit python

miurahr:
by running `tox -e py39d`

```
================================ leaks summary =================================
tests/test_install.py::test_install[cmd0-linux-desktop-1.2.3-0-197001020304---linux_x64/desktop/tools_qtcreator/Updates.xml-archives0-^aqtinstall\\(aqt\\) v.* on Python 3.*\\nDownloading qt.tools.qtcreator...\\nFinished installation of tools_qtcreator-linux-qt.tools.qtcreator.7z in .*\\nFinished installation\\nTime elapsed: .* second]: leaked references: [42, 43, 44, 45], memory blocks: [16, 16, 16, 16]
tests/test_install.py::test_install[cmd1-linux-desktop-1.2.3---linux_x64/desktop/tools_qtcreator/Updates.xml-archives1-^aqtinstall\\(aqt\\) v.* on Python 3.*\\nWarning: The command 'tool' is deprecated and marked for removal in a future version of aqt.\\nIn the future, please use the command 'install-tool' instead.\\nDownloading qt.tools.qtcreator...\\nFinished installation of tools_qtcreator-linux-qt.tools.qtcreator.7z in .*\\nFinished installation\\nTime elapsed: .* second]: leaked references: [52, 53, 54, 55], memory blocks: [16, 16, 16, 16]
tests/test_install.py::test_install[cmd2-windows-desktop-5.14.0-win32_mingw73-mingw73_32-windows_x86/desktop/qt5_5140/Updates.xml-archives2-^aqtinstall\\(aqt\\) v.* on Python 3.*\\nWarning: The command 'install' is deprecated and marked for removal in a future version of aqt.\\nIn the future, please use the command 'install-qt' instead.\\nDownloading qtbase...\\nFinished installation of qtbase-windows-win32_mingw73.7z in .*\\nFinished installation\\nTime elapsed: .* second]: leaked references: [62, 63, 64, 65], memory blocks: [16, 16, 16, 16]
[...]

PR #465, opened 2021-12-07
Cache PowerShell Modules in Azure Pipelines

ddalcino:
There are frequent failures in Azure Pipelines builds for Windows, caused by failures to retrieve the the Powershell modules 'Pscx' and 'VSSetup' from the PowerShell Gallery. This PR is intended to cache the Powershell modules required by the Azure pipelines, so that these failures stop occurring. As a side effect, it will also speed up all Windows build jobs by however long it takes to download the powershell modules (expect 1-2 minutes).

Examples of failures:
https://dev.azure.com/miurahr/github/_build/results?buildId=4817&view=logs&j=9f3e9a58-f4d0-59eb-bb9a-f7c4195515fd&t=d65f8022-fa96-505b-389b-6aa2d54135b0&l=27 (from #459)
https://dev.azure.com/miurahr/github/_build/results?buildId=4831&view=logs&j=dcd50a69-7bd8-59f7-72b1-a119e6ffb32f&t=b0305de3-bc12-557b-67f0-1215b8f079d0&l=28 (from #464)

Citation: https://blog.simonw.se/caching-powershell-modules-in-azure-pipelines/

### DIFF
--- a/ci/steps.yml
+++ b/ci/steps.yml
@@ -179,9 +179,6 @@ steps:
   - task: Cache@2
     inputs:
       key: '"pwsh modules" | "$(startYear)" | "$(startMonth)"'
-      restorekeys: |
-        "pwsh modules" | "$(startYear)"
-        "pwsh modules"
       path: $(MODULES_FOLDER)
       cacheHitVar: PSModules_IsCached
     condition: |
--- a/ci/steps.yml
+++ b/ci/steps.yml
@@ -175,13 +175,35 @@ steps:
     displayName: Build accelbubble example application to test for android
 
   ##----------------------------------------------------
-  # determine Windows build system
+  # Cache Powershell Modules in $MODULES_FOLDER
+  - task: Cache@2
+    inputs:
+      key: '"pwsh modules" | "$(startYear)" | "$(startMonth)"'
+      restorekeys: |
+        "pwsh modules" | "$(startYear)"
+        "pwsh modules"
+      path: $(MODULES_FOLDER)
+      cacheHitVar: PSModules_IsCached
+    condition: |
+        and(eq(variables['Agent.OS'], 'Windows_NT'), 
+            eq(variables['SUBCOMMAND'], 'install-qt'))
+    displayName: Cache Powershell Modules
+  # On cache miss: Download Powershell Modules to $MODULES_FOLDER
   - powershell: |
       Install-PackageProvider NuGet -Force
       Import-PackageProvider NuGet -Force
       Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
-      Install-Module Pscx -AllowClobber -AllowPrerelease
-      Install-Module VSSetup -Scope CurrentUser
+      md -p $(MODULES_FOLDER)   # mkdir
+      Save-Module -Name Pscx -AllowPrerelease -Path $(MODULES_FOLDER)
+    condition: |
+        and(eq(variables['Agent.OS'], 'Windows_NT'), 
+            eq(variables['SUBCOMMAND'], 'install-qt'),
+            ne(variables.PSModules_IsCached, 'true'))
+    displayName: Download Powershell Modules for Windows on cache miss
+
+  ##----------------------------------------------------
+  # determine Windows build system
+  - powershell: |
       if ('$(ARCH)' -like '*msvc*') {
         Write-Host '##vso[task.setvariable variable=TOOLCHAIN]MSVC'
       }
@@ -202,7 +224,9 @@ steps:
       }
       cd $(WIN_QT_BINDIR)
       unzip $(Build.SourcesDirectory)\ci\jom_1_1_3.zip
-    condition: and(eq( variables['Agent.OS'], 'Windows_NT'), eq(variables['SUBCOMMAND'], 'install-qt'))
+    condition: |
+        and(eq(variables['Agent.OS'], 'Windows_NT'), 
+            eq(variables['SUBCOMMAND'], 'install-qt'))
     displayName: Detect toolchain for Windows and update PATH
 
   # When no modules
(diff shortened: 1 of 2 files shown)


ci/steps.yml
```diff
@@ -217,6 +239,12 @@ steps:
     displayName: Build test with qmake for Linux and macOS w/o extra module
   - powershell: |
       if ( $env:TOOLCHAIN -eq 'MSVC' ) {
+        # Load modules from cache  
+        $Env:PSModulePath = '$(MODULES_FOLDER)', $Env:PSModulePath -join [System.IO.Path]::PathSeparator
+        Write-Host $Env:PSModulePath
+        Import-Module "Pscx"
+        Import-Module "VSSetup"
+
```
ddalcino, line 247: There is a DRY issue here: This block of code is repeated on lines 275-280. I think it would make a lot more sense to run this code in a previous step, but I do not know how to make the effect of this code persist to future steps. I suspect that there's a way to make this work without repeating this code, but I don't know how.

ci/steps.yml
```diff
@@ -175,13 +175,33 @@ steps:
     displayName: Build accelbubble example application to test for android
 
   ##----------------------------------------------------
-  # determine Windows build system
+  # Cache Powershell Modules in $MODULES_FOLDER
+  - task: Cache@2
+    inputs:
+      key: 'pwsh modules | "$(Agent.OS)" | Pscx | VSSetup'
```
ddalcino, line 181: This key will need to change in the future if we need to download other modules, or if these modules are updated. If we don't change the key, Azure will just load the same cached copies over and over again.

It would probably make sense to add the version numbers for the Pscx and VSSetup modules to this key, but I don't know how to do that.

If we can safely assume that we won't ever need updated versions of these modules, maybe this implementation is OK.

ci/steps.yml
miurahr: We can define a custom shell  for pipeline that based is pwsh and preloading modules,  but it would be overkill.

ci/steps.yml
miurahr: When Microsoft releases new version of Visual Studio such as VS2022,  and we want to test with it for sample project, we may want to update these modules to recognize new VS version.

(cf. https://github.com/Pscx/Pscx/pull/89)

ddalcino: That's a good point. 

One reasonable option would be to lock the key to the `pool.vmImage` property, which is `windows-2019`. When we move up to a new windows image that includes VS2022, we would cache new versions of the modules.

ddalcino:
According to [Azure windows-2019 docs](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#powershell-modules), `VSSetup` v2.2.16 is already installed on the `windows-2019` image! Do we need to do any of this install/cache stuff at all?

ci/steps.yml
```diff
@@ -175,13 +175,34 @@ steps:
     displayName: Build accelbubble example application to test for android
 
   ##----------------------------------------------------
-  # determine Windows build system
+  # Cache Powershell Modules in $MODULES_FOLDER
+  - task: Cache@2
+    inputs:
+      key: '"pwsh modules" | "$(Agent.OS)" | "$(Agent.Name)" | "$(Agent.Version)"'
+      # Agent.Version will invalidate the cache every few weeks, when a new agent is released
```
ddalcino, line 182: `"$(Agent.Name)" | "$(Agent.Version)"` are the wrong properties to use here. I was hoping to use a variable that evaluates to `windows-2019`, and a version number for the VM image, but I have not found one.

miurahr:
> ~According to [Azure windows-2019 docs](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#powershell-modules), `VSSetup` v2.2.16 is already installed on the `windows-2019` image! Do we need to do any of this install/cache PS modules stuff at all?~
> 
> `Import-VisualStudioVars` is part of Pscx, which is not installed in the Win-2019 image. Caching the modules will be necessary.

`Import-VisualStudioVars` function is provided by `Pscx` that use `VSSetup` to detect a install path of Visual  Studio, and it invoke batch file `vcvarsall.bat` or `VsDevCmd.bat` to set environment variable to power shell session.
We can do it by ourselves, in theory, and remove a dependency to `Pscx`.

Path and name of batch file is vary on host architectures, vs versions, and target arch, such as 
 `$installPath/Common7/Tools/x86/x64/VsDevCmd.bat`

ref https://github.com/Pscx/Pscx/blob/440990bf025483349acb998e14351e8289af1b5e/Src/Pscx/Modules/Utility/Pscx.Utility.psm1#L2111

ddalcino:
> `Import-VisualStudioVars` function is provided by `Pscx` that use `VSSetup` to detect a install path of Visual Studio, and it invoke batch file `vcvarsall.bat` or `VsDevCmd.bat` to set environment variable to power shell session. We can do it by ourselves, in theory, and remove a dependency to `Pscx`.

If that's all `Import-VisualStudioVars` does, then I think that it will be unlikely that `Pscx` will change much while we are using a single image of Windows. To my knowledge, 'windows-2019' will always include VS-2019, and 'windows-2022' will always include VS-2022. I think it makes sense to cache one copy of `Pscx` per Windows image, to account for changes to the module when a new version of Visual Studio is released. If we start seeing build failures because `Pscx` is too old, we can revisit this.

This isn't a perfect solution, but I think it will work well enough for now.

ci/steps.yml
```diff
@@ -175,13 +175,33 @@ steps:
...
+  # On cache miss: Download Powershell Modules to $MODULES_FOLDER
   - powershell: |
       Install-PackageProvider NuGet -Force
       Import-PackageProvider NuGet -Force
       Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
-      Install-Module Pscx -AllowClobber -AllowPrerelease
-      Install-Module VSSetup -Scope CurrentUser
+      md -p $(MODULES_FOLDER)   # mkdir
+      Save-Module -Name Pscx -AllowPrerelease -Path $(MODULES_FOLDER)
+      Save-Module -Name VSSetup -Path $(MODULES_FOLDER)
```
ddalcino, line 195:
```suggestion
```
Maybe this part should be removed, since VSSetup should already exist in the `windows-2019` image, and will often be more up-to date than our cached copy?

ci/steps.yml
```diff
@@ -175,13 +175,33 @@ steps:
     displayName: Build accelbubble example application to test for android
 
   ##----------------------------------------------------
-  # determine Windows build system
+  # Cache Powershell Modules in $MODULES_FOLDER
+  - task: Cache@2
+    inputs:
+      key: '"pwsh modules" | "$(WINDOWS_VM_IMAGE)"'
```
ddalcino, line 181: Maybe it would make more sense here to use the date instead of the name of the image? Something like:
```suggestion
      key: '"pwsh modules" | "$[format('{0:MMyyyy}', pipeline.startTime)]"'
```
This would invalidate the cache once at the beginning of every month, so our copy of Pscx could be at most one month behind the latest version.

Citation for date formatting: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#format

I'm not too sure about the quoting/escaping of quote characters here.

PR #466, opened 2021-12-08
Azure-Win-2019: use built-in VSSetup module

ddalcino:
Not sure this is going to work, but it's worth trying.


ddalcino:
I didn't realize that Import-VisualStudioVars was part of Pscx, not VSSetup. This will not work.

PR #471, opened 2021-12-29
Add 32 bit binary to release

lebarsfa:
Here is a proposition to release a 32 bit binary (there are still computers of less than 5 years old limited to 32 bit Windows, e.g. Intel Compute Stick, ZOTAC Pico, ASUS Transformer). This would probably simplify also the installation of this [Chocolatey package](https://community.chocolatey.org/packages/qt5-default) by removing the Python dependency.
Separate binaries `aqt_x64.exe` and `aqt_x86.exe` are generated, see https://github.com/lebarsfa/aqtinstall/releases. I do not know if the continuous release `aqt.exe` should be in 64 or 32, at the moment it is left to 64 in that code.

### DIFF
--- a/.github/workflows/upload-release-artifacts.yml
+++ b/.github/workflows/upload-release-artifacts.yml
@@ -17,6 +17,7 @@ jobs:
       matrix:
         os: [windows-latest]
         py: [3.9]
+        arch: [x86, x64]
     steps:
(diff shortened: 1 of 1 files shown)


miurahr:
Thank you for the proposal and contribution.

Issue #472, opened 2022-01-09
qmake does not work when qtlocation failed by ArchiveConnectionError during installation process


martinrotter:
```
Finished installation of d3dcompiler_47-x64.7z in 1.21278810
Finished installation of qtbase-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z in 8.40442290
Caught ArchiveConnectionError, terminating installer workers
Failure to connect to https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt5_5152/qt.qt5.5152.win64_msvc2019_64/5.15.2-0-202011130602qtlocation-Windows-Windows_10-MSVC2019-Windows-Windows_10-X86_64.7z.sha1: ReadTimeout
```

- aqtinstall(aqt) v2.0.5 on Python 3.7.9 [CPython MSC v.1900 64 bit (AMD64)]
- this command used - https://github.com/martinrotter/rssguard/blob/master/resources/scripts/github-actions/build-windows.ps1#L47

It seems it fails to download some packages, also, I am now seeing this error when trying to use "qmake":

```
Error: Qt license file was not found!
Project ERROR: License check failed! Giving up ...
```

ddalcino:
Try repeating the aqt install-qt command. That’s an HTTP error, it just means aqt failed to communicate with the server. This is normal.

ddalcino:
It's really hard to debug this without a stack trace. @martinrotter, could you provide complete program output, as requested by our [issue template](https://github.com/miurahr/aqtinstall/blob/master/.github/ISSUE_TEMPLATE/bug_report.md)?

I am guessing here, but I think the root cause of this is this line:
https://github.com/miurahr/aqtinstall/blob/2f328bde8abe9da9521f616bc597671e7903f8fd/aqt/installer.py#L1023

The `getUrl` call above is asking for a `.sha1` hash file. I think that what's happening here is that `getUrl` is calling `requests.get`, which raises `requests.exceptions.Timeout`, which gets wrapped up as an `ArchiveConnectionError`.
https://github.com/miurahr/aqtinstall/blob/2f328bde8abe9da9521f616bc597671e7903f8fd/aqt/helper.py#L52-L60

I'm not sure about this, but to me it looks like the call to `requests.get` is not being retried. @miurahr, is line 60 supposed to be `session.get`?
```diff
- r = requests.get(url, allow_redirects=False, timeout=timeout)
[...]

PR #473, opened 2022-01-11
Fix behavior that getUrl() does not retry

miurahr:
getUrl() now use `requests.get()` that is default getter that does not retry.
This change corrects it to `session.get()`

Related issue #472


miurahr:
tests failed because test mock is incomplete against the change.

PR #485, opened 2022-02-06
Update fallback mirror list

hoffie:
Some of the currently listed mirrors are no longer part of the official mirror list by qt.io, some return 404, some don't fully mirror the tree (especially `online/`).

The updated list has been generated via the following one-liner:
```bash
for mirror in $(curl -s https://download.qt.io/static/mirrorlist/ | grep -oP 'href="\K(http.*)(?=">HTTP</a>)' | sed -re 's|^http:|https:|; s|([^/])$|\1/|'); do code=$(curl -I "${mirror}online/qtsdkrepository/windows_x86/desktop/qt6_623/qt.qt6.623.win64_msvc2019_64/6.2.3-0-202201260729qtbase-Windows-Windows_10_21H2-MSVC2019-Windows-Windows_10_21H2-X86_64.7z" --write-out '%{response_code}' --silent -o /dev/null); [[ $code == 200 ]] && echo "    ${mirror}"; done
```

### DIFF
--- a/aqt/settings.ini
+++ b/aqt/settings.ini
@@ -23,21 +23,20 @@ blacklist:
     http://mirrors.geekpie.club
 fallbacks:
     https://qtproject.mirror.liquidtelecom.com/
+    https://mirrors.aliyun.com/qt/
     https://mirrors.sjtug.sjtu.edu.cn/qt/
     https://mirrors.ustc.edu.cn/qtproject/
     https://ftp.jaist.ac.jp/pub/qtproject/
     https://ftp.yz.yamagata-u.ac.jp/pub/qtproject/
     https://qt-mirror.dannhauer.de/
     https://ftp.fau.de/qtproject/
+    https://mirror.netcologne.de/qtproject/
     https://mirrors.dotsrc.org/qtproject/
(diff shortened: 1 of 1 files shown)


aqt/settings.ini
```diff
@@ -23,22 +23,23 @@ blacklist:
     http://mirrors.geekpie.club
 fallbacks:
     https://qtproject.mirror.liquidtelecom.com/
+    https://mirrors.aliyun.com/qt/
     https://mirrors.sjtug.sjtu.edu.cn/qt/
+    https://mirrors.tuna.tsinghua.edu.cn/qt/
```
miurahr, line 28: This is problematic because it limit access connections unknown condition, maybe outside of china, and randomly failed, So it is listed in black list and it had been failed in past CI tests.
Could you remove blacklisted host?

miurahr: Or if  it is healthy, please also update blacklist.

hoffie: Thanks for the pointer, should have caught that myself. I've updated the PR to avoid adding blacklisted entries.
While the mirror is currently reachable for me, it's rather slow and I don't know the history, so it might be better to keep it as-is.

Issue #488, opened 2022-02-07
Feature: Try multiple fallbacks on failure


ann0see:
**Is your feature request related to a problem? Please describe.**
Sometimes our CI fails due to download failures from Qt mirrors. Until yesterday, we didn't know aqt already polls multiple mirrors. Strangely enough this is not always sufficient to have a successful download if multiple mirrors are down. 

**Describe the solution you'd like**
Automatically try 3 or more mirrors before failing.
What does the code actually do at the moment? I thought it already tries multiple mirrors.

**Describe alternatives you've considered**
* Add a ci option to specify the number of mirrors being checked before aqt fails.
* Handle download failures outside of aqt (that's what we're trying at the moment).

**Additional context**
See my PR on our repo: https://github.com/jamulussoftware/jamulus/pull/2333 which is supposed to work around download failures and https://github.com/jamulussoftware/jamulus/runs/5040398208?check_suite_focus=true, https://github.com/jamulussoftware/jamulus/runs/5038967837?check_suite_focus=true
https://github.com/jamulussoftware/jamulus/runs/4966741224?check_suite_focus=true

https://github.com/jamulussoftware/jamulus/runs/4857453378?check_suite_focus=true (seems unrelated)

ann0see:
@hoffie after looking at the code I think it should try multiple mirrors already...

hoffie:
Hm, do you have a pointer? I thought it would try only a single fallback and I thought this was the issue.
https://github.com/miurahr/aqtinstall/blob/eaf5a46e488b69ec1503ca0695020bd7f084ae66/aqt/installer.py#L920

ann0see:
Maybe it's about some master mirror failing?

ddalcino:
This particular failure:
> https://github.com/jamulussoftware/jamulus/runs/4857453378?check_suite_focus=true (seems unrelated)

... actually failed right here: https://github.com/jamulussoftware/jamulus/runs/4857453378?check_suite_focus=true#step:4:148. I don't know why the nonzero return code didn't stop that script, but I am sure that this is the same issue as the other failures you linked to.

The failures you are having are all occurring when aqtinstall is trying to download a `.sha1` file. The code that downloads this file is right here:
https://github.com/miurahr/aqtinstall/blob/cb7eaeeb2afb755a3020ad510b74e62647b3ab99/aqt/installer.py#L1024
[...]

PR #491, opened 2022-02-19
Add explanation of --config flag in CLI docs

ddalcino:
This adds the `-c | --config` flag to the CLI section of the documentation.

This problem was discovered while investigating #488. It does not attempt to fix that issue.


ann0see (commented):
Good to know about the CLI argument.

miurahr:
Thank you for the fix. merged.

PR #507, opened 2022-03-27
Support pyproject.toml:project section

miurahr:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,8 @@ Requirements
     py7zr
     texttable
     bs4
-    dataclasses
+    dataclasses; python_version < "3.7"
+    defusedxml
 
 - Operating Systems:
     Linux, macOS, MS Windows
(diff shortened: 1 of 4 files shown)


miurahr:
@nikitalita When adding [project] section in `pyproject.toml' execution of standalone binary failed. Could you investigate it?

nikitalita:
It looks like it doesn’t like 
```

 [project.scripts]
 aqt = "aqt.__main___:main"

````
This has to point to the actual Mai no function, I think?

**Comment by miurahr** [comment]
> It looks like it doesn’t like
> 
> ```
> 
>  [project.scripts]
>  aqt = "aqt.__main___:main"
> ```
> 
> This has to point to the actual main function, I think?

Thank you for the suggestion.
I'd like to add config as same as one in setup.cfg;

```
[options.entry_points]
console_scripts =
    aqt = aqt.__main__:main
```
I saw https://peps.python.org/pep-0621/#entry-points

PR #510, opened 2022-03-28
Fix patching of Qt6.2.2-ios

### LINKED ISSUE #503: ios: qmake script contains non-resolvable target path for Qt 6.2.x
**Describe the bug**
Qt 6.2.x (maybe earlier versions as well) seems to have changed regarding the `ios` architecture compared to 5.15.2:
1, `$OUTPUT_DIR/6.2.3/ios/bin/qmake` is no longer independent, instead, the `macos/desktop` architecture has to be installed additionally. That's fine (although a bit unintuitive, but I guess there is not much which can be done about it). Maybe aqt could warn about that or even install the required archive (macos/desktop/qtbase) automatically?
2. It looks like `$OUTPUT_DIR/6.2.3/ios/bin/qmake` is supposed to find the `macos` `desktop` qmake, but currently it doesn't due to path changes.

#289 seems related, but apparently isn't enough to fix this.

**To Reproduce**
1. `python3 -m pip install "aqtinstall==2.0.6"`
2. `python3 -m aqt install-qt --outputdir "/usr/local/opt/qt/" mac ios "6.2.3" --archives qtbase qttools qttranslations`
3. `/usr/local/opt/qt/6.2.3/ios/bin/qmake`

**Expected behavior**
qmake only works when manually creating a symlink from the expected location to the new location like that:
https://github.com/hoffie/jamulus/commit/6c7e4751a1b5461e62e17600e2047919a2ce7f4f#diff-ad84ad4ff36c1685f58e8d153980cb299a85fd0f851681459248564f4d3a9b13R27
Another workaround would be not calling ios/bin/qmake by users, but if this is the advise, then ios/bin/qmake should probably not be there in the first place. :)

aqt should insert the proper path into the `qmake` script.

I believe that this place has to be adjusted with version-specific logic:
https://github.com/miurahr/aqtinstall/blob/1a82a92e4c44b0891b6df03f1f5ff371eff2de50/aqt/updater.py#L176

`clang_64` should be `macos` for Qt 6.2.x.
I may try to submit a PR this evening (but feel free to go ahead if you don't hear from me).

**`aqt` output**
```
/usr/local/opt/qt/6.2.3/ios/bin/qmake: line 7: /usr/local/opt/qt/6.2.3/clang_64/bin/qmake: No such file or directory
```

 - macOS 10.15 with ios target
 - `aqtinstall(aqt) v2.0.6 on Python 3.9.10 [CPython Clang 12.0.0 (clang-1200.0.32.29)]` (Update: 2.1.0rc2 shows the same behavior)

**Additional context**
Of course, I might be missing something; I've just noticed that simply replacing 5.15.2 by 6.2.3 no longer leads to a working `qmake` setup for `ios`.

ddalcino:
Fix #503.

~Disregard for now; I just want to see 2 new tests that fail.~ This adds tests and fixes that allow the tests to pass.

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -272,7 +272,7 @@ class Updater:
                 arch_dir = b + "_" + a
             else:
                 arch_dir = arch[6:]
-        elif os_name == "mac" and target.target == "desktop":
+        elif os_name == "mac" and arch == "clang_64":
             arch_dir = default_desktop_arch_dir(os_name, version)
         else:
             arch_dir = arch
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -24,6 +24,7 @@ import pathlib
 import re
 import subprocess
 from logging import getLogger
+from typing import Union
 
 import patch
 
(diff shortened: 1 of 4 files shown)


ddalcino:
We have a problem: I expected https://github.com/miurahr/aqtinstall/pull/510/checks?check_run_id=571360059 to fail, but it does not, because it does not attempt to build a qt iOS project. We will need to add one before I can start working on an implementation.

aqt/updater.py
```diff
@@ -274,8 +272,8 @@ def update(cls, target: TargetConfig, base_dir: str):
                 arch_dir = b + "_" + a
             else:
                 arch_dir = arch[6:]
-        elif version in SimpleSpec(">=6.1.2") and os_name == "mac" and arch == "clang_64":
-            arch_dir = "macos"
+        elif os_name == "mac" and target.target == "desktop":
+            arch_dir = default_desktop_arch_dir(os_name, version)
```
ddalcino, line 276: This change is inappropriate because it breaks the macos/wasm_32 build. Perhaps this aggressive refactoring was a mistake.

miurahr:
Thank you. Squashed and merged into master.

PR #518, opened 2022-04-20
Allow MetadataFactory to recover from bad HTML files

ddalcino:
Fix #517

This change allows MetadataFactory.iterate_folders to recognize when MetadataFactory has retrieved a malformed HTML file, or an HTML file that does not contain the table and rows that it expects to see.

If the URL is correct, it should retrieve the right page, but this is not guaranteed, and MetadataFactory should be able to recover and generate an appropriate error message.

One possible situation where this kind of error might occur could be when the user is connected to a public WIFI network, but not logged in. The router would redirect aqt to the network login page, and aqt would have no way of knowing that it did not receive the correct page. Aqt should be resilient to this kind of error.

Issue #517, opened 2022-04-20
`MetadataFactory` makes inappropriate assumptions about the HTML files it retrieves.


ddalcino:
**Describe the bug**
`MetadataFactory` improperly assumes that HTML pages at download.qt.io will always contain a body that contains a table. Infrequently, the server will return something unexpected, and `aqt` should be able to provide a reasonable error message.

**To Reproduce**
Run `aqt list-qt linux desktop`, under the condition that the URL https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/ returns a page without a table.

**Expected behavior**
The `aqt list-qt linux desktop` command returns a list of Qt versions.

**`aqt` output**

```
'NoneType' object has no attribute 'find_all'
Traceback (most recent call last):
File "/home/garald/.local/lib/python3.8/site-packages/aqt/installer.py", line 108, in run
args.func(args)
File "/home/garald/.local/lib/python3.8/site-packages/aqt/installer.py", line 524, in run_list_qt
show_list(meta)
File "/home/garald/.local/lib/python3.8/site-packages/aqt/metadata.py", line 775, in show_list
output = meta.getList()
File "/home/garald/.local/lib/python3.8/site-packages/aqt/metadata.py", line 435, in getList
return self._action()
File "/home/garald/.local/lib/python3.8/site-packages/aqt/metadata.py", line 473, in fetch_versions
versions = sorted(filter(None, map(get_version, filter(filter_by, versions_extensions))))
File "/home/garald/.local/lib/python3.8/site-packages/aqt/metadata.py", line 600, in iterate_folders
for row in soup.body.table.find_all("tr"):
AttributeError: 'NoneType' object has no attribute 'find_all'
aqtinstall(aqt) v2.1.0 on Python 3.8.10 [CPython GCC 9.4.0]
Working dir: /home/garald
Arguments: ['/home/garald/.local/bin/aqt', 'list-qt', 'linux', 'desktop'] Host: uname_result(system='Linux', node='garald-H310M-S2-2-0', release='5.4.0-109-generic', version='#123-Ubuntu SMP Fri Apr 8 09:10:54 UTC 2022', machine='x86_64', processor='x86_64')
===========================PLEASE FILE A BUG REPORT===========================
You have discovered a bug in aqt.
Please file a bug report at https://github.com/miurahr/aqtinstall/issues.
Please remember to include a copy of this program's output in your report.
```

**Desktop (please complete the following information):**
 - OS: Linux
[...]

PR #520, opened 2022-05-05
Properly retrieve folder names from html pages of all mirrors

### LINKED ISSUE #519: aqtinstall does not work when https://download.qt.io is not available
Hello! When https://download.qt.io is not available it is not possible to use mirrors from the settings.ini config to list available tools:

```
$ aqt list-tool linux desktop
'NoneType' object has no attribute 'find_all'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/aqt/installer.py", line 108, in run
    args.func(args)
  File "/usr/local/lib/python3.8/dist-packages/aqt/installer.py", line 540, in run_list_tool
    show_list(meta)
  File "/usr/local/lib/python3.8/dist-packages/aqt/metadata.py", line 775, in show_list
    output = meta.getList()
  File "/usr/local/lib/python3.8/dist-packages/aqt/metadata.py", line 435, in getList
    return self._action()
  File "/usr/local/lib/python3.8/dist-packages/aqt/metadata.py", line 482, in fetch_tools
    return list(MetadataFactory.iterate_folders(html_doc, "tools"))
  File "/usr/local/lib/python3.8/dist-packages/aqt/metadata.py", line 600, in iterate_folders
    for row in soup.body.table.find_all("tr"):
AttributeError: 'NoneType' object has no attribute 'find_all'
aqtinstall(aqt) v2.1.0 on Python 3.8.10 [CPython GCC 9.4.0]
Working dir: `/home/user`
Arguments: `['/usr/local/bin/aqt', 'list-tool', 'linux', 'desktop']` Host: `uname_result(system='Linux', node='laptop', release='5.13.0-40-generic', version='#45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022', machine='x86_64', processor='x86_64')`
===========================PLEASE FILE A BUG REPORT===========================
You have discovered a bug in aqt.
Please file a bug report at https://github.com/miurahr/aqtinstall/issues.
Please remember to include a copy of this program's output in your report.
```

Is it possible to use aqtinstall with mirrors when https://download.qt.io is not available?

mite-user:
When https://download.qt.io is not available, `aqt list-*` command falls back to a random mirror.
Different mirrors provide different html layouts.

Mirrors with `<table>` similar enough to be properly handled by the current script:
https://download.qt.io/
https://ftp.jaist.ac.jp/pub/qtproject/
https://qt-mirror.dannhauer.de/
https://master.qt.io/

Mirrors with `<table>` which isn't properly handled by the current script:
https://mirrors.aliyun.com/qt/
https://mirrors.sjtug.sjtu.edu.cn/qt/
https://mirrors.dotsrc.org/qtproject/

Mirrors with a list of `<a>` elements inside `<pre>`, no `<table>`:
https://qtproject.mirror.liquidtelecom.com/
https://mirrors.ustc.edu.cn/qtproject/
https://ftp.yz.yamagata-u.ac.jp/pub/qtproject/
https://ftp.fau.de/qtproject/
https://mirror.netcologne.de/qtproject/
https://www.nic.funet.fi/pub/mirrors/download.qt-project.org/
https://mirrors.ukfast.co.uk/sites/qt.io/
https://qt.mirror.constant.com/

Mirrors with a list of `<a>` elements inside `<pre>`, with `<table>` which doesn't contain folders:
https://ftp2.nluug.nl/languages/qt/
https://ftp1.nluug.nl/languages/qt/

If the mirror has similar enough html layout, the program works and prints the requested info.

If there is no `<table>` element,
v2.1.0 fails with the error described in #519 ;
the master branch prints
```
Failed to retrieve the expected HTML page at https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/
```
This is an unrelated bug. The program should be printing the fallback url, not baseurl.

If there is a `<table>` element, the program prints
```
No versions available for this request.
```
or
```
No tools available for this request.
```

## Solution.
Get folder names from value of `href` attribute of `<a>` elements.
Folder links point to a page one level deeper relative to the current page,
i.e. the value should look like `./foo` or `foo`,
NOT `https://foo`, `../foo`, `..`, `/foo`, `?foo`.

## How to test.
```
git clone https://github.com/mite-user/aqtinstall.git
echo '__version__ = "handle-html-test"' > aqtinstall/aqt/version.py
```
Open `aqtinstall/aqt/settings.ini`,
set `baseurl:` to `http://localhost`,
remove all mirrors expect the one to be tested from `fallbacks:`.
```
cd aqtinstall
python3 -m aqt list-qt linux desktop
```

This commit might fix #519 .


aqt/metadata.py
```diff
@@ -589,20 +589,26 @@ def fetch_http(rest_of_url: str, is_check_hash: bool = True) -> str:
                     )
 
     def iterate_folders(self, html_doc: str, filter_category: str = "") -> Generator[str, None, None]:
-        def table_row_to_folder(tr: bs4.element.Tag) -> str:
+        def link_to_folder(link: bs4.element.Tag) -> str:
             try:
-                return tr.find_all("td")[1].a.contents[0].rstrip("/")
+                url: str = link.get("href")
+                url = url.split("?")[0].rstrip("/")
```
ddalcino, line 595: Have you looked at `urllib.parse`? If you import `urlparse` from `urllib.parse`, you can do this:
```suggestion
                url = urlparse(url).path.rstrip("/")
```
... and get the benefit of a thoughtfully-designed URL parsing library that will take out anything else that doesn't belong in the path (for instance, `#some_id`).

I'm not sure we really need to use `urlparse`, but I think it's worth asking.

aqt/metadata.py
```diff
@@ -589,20 +589,26 @@ def fetch_http(rest_of_url: str, is_check_hash: bool = True) -> str:
...
             except (AttributeError, IndexError):
                 return ""
 
         try:
             soup: bs4.BeautifulSoup = bs4.BeautifulSoup(html_doc, "html.parser")
-            for row in soup.body.table.find_all("tr"):
-                content: str = table_row_to_folder(row)
-                if not content or content == "Parent Directory":
+            for link in soup.find_all("a"):
+                folder: str = link_to_folder(link)
```
ddalcino, line 607: Thank you for renaming this; its much easier to read now.

ddalcino (changes requested):
This looks good, and it's a welcome improvement! Thank you.

The only real suggestion I have is that you add a test or two that the existing implementation cannot pass. In your discussion of the problem, you mentioned several different cases where the mirror HTML pages have a different structure than the pages at download.qt.io. I would like to see tests that run `iterate_folders` or `get_versions_extensions` on a couple of examples of these differently-structured HTML files, and expect the correct set of results. This will help us to prevent regressions in the future.

mite-user:
### Updated the original commit
1. Using `urllib.parse.urlparse`.
2. Handling `./` at the beginning and `/` at the end with `posixpath.normpath`.
3. Checking for one and two dots only. Apparently three and more dots point one level deeper.



### Second commit
Added test html files for `test_list_versions_tools()` in `tests/test_list.py`.



### Notes
Absence of `<table>` element no longer triggers the error added in #518 . There should be some other way to check if html contents are unexpected. `test_list_bad_html()` now always fails.

`list-tool` doesn't sort the resulting list. So the order depends on what is provided by html. Looks like all or almost all mirrors provide sorted links by default. They are unsorted at https://download.qt.io .

aqt/metadata.py
```diff
@@ -24,6 +24,7 @@
 import re
 import secrets as random
 import shutil
+from urllib.parse import urlparse
```
ddalcino, line 27: This part of the 'check' action: https://github.com/miurahr/aqtinstall/runs/6331277720?check_suite_focus=true#step:7:17 failed because the imports are no longer sorted. You can fix that with isort.

If this helps, I always run `isort aqt tests && black aqt tests` before I commit anything that's going into this codebase, just to avoid any code formatting or import sorting issues. I think you can fix them with `flake8` as well.

aqt/metadata.py
```diff
@@ -26,6 +26,7 @@
 import shutil
 from logging import getLogger
 from typing import Callable, Dict, Generator, Iterable, Iterator, List, Optional, Tuple, Union
+from urllib.parse import ParseResult, urlparse
 from xml.etree.ElementTree import Element
```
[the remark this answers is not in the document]
miurahr: We have already use defusedxml as dependency. Please feel free to use it.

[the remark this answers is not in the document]
mite-user: I think it's complaining about `from xml.etree.ElementTree import Element`. It's irrelevant to this pull request.
Not an issue, apparently.
https://github.com/miurahr/aqtinstall/pull/498#discussion_r824344337

mite-user:
### Changed `link_to_folder()`.

Removed `try:` `except (AttributeError, IndexError):`.
With the new code, exceptions are unexpected and shouldn't be ignored.

Changed `link.get("href")` to `link.get("href", default="")`.
Without `default=""`, it will return `None` instead of empty `str` if there is no `href` in `<a>`.

ddalcino:
> Absence of `<table>` element no longer triggers the error added in #518 . There should be some other way to check if html contents are unexpected. `test_list_bad_html()` now always fails. 

For the record, I’d like to say that I support removing the `test_list_bad_html()` test that now fails. I don't think that it really adds all that much. 

I wrote that test under the assumption that #517 was caused by being in a coffee shop, or by some http error. Given the evidence that @mite-user presents here, l think that it's much more likely that a different html structure in a mirror site was the cause. The requests library should be able to detect serious HTTP errors that would result in malformed html, and I don't think you get a 200 status code when a router asks you to sign in.

ddalcino:
@miurahr Any further thoughts on this? I still think we should remove `test_list_bad_html()` and merge this. I don't think there's any obvious way to merge this without removing that test, and I don't think the test is useful.

miurahr:
@ddalcino 
>  I still think we should remove test_list_bad_html() and merge this. 

I think it is OK to remove and merge this.

mite-user:
Should I add a commit to this pull request to remove `test_list_bad_html()`?

ddalcino:
> Should I add a commit to this pull request to remove `test_list_bad_html()`?

IMHO, yes, you should add a commit that removes the failing test. I don’t think we need a separate PR for this. If we did, I think we would have heard from @miurahr by now.

This PR effectively replaces #518, which is where `test_list_bad_html()` comes from. Speaking as the author of that test, I think the right way to replace it is to just remove it.

ddalcino (approved):
All tests are passing, and the important parts of this code have test coverage. It looks like `aqt/metadata.py:611-613` is no longer covered, but I don't think it matters. IMHO this is ready to merge!

PR #540, opened 2022-07-24
Automatically install desktop qt when required for android/ios qt installations 

### LINKED ISSUE #528: Android installs missing gcc_64 directory
If you install Android Qt6 builds on linux like (this is how the github action works):
`python3 -m aqt install 6.3.0 linux android android_armv7 -O /home/runner/work/GPXSee/Qt`
the resulting Qt installation is broken as it is missing the gcc_64 directory which contains the actual
Qt binaries (in android_armv7 there are only symlinks) like qmake.

As a result, qmake-based builds are not possible with Qt installed using aqtinstall.

ddalcino:
I think this should fix #528.

* This adds a CLI flag, `--autodesktop`, that automatically installs the required desktop version of Qt if it is not already installed. I chose not to turn it on by default, in order to preserve the existing behavior.
* This adds a warning when `--autodesktop` is not turned on and the expected desktop version of Qt is not present. The warning prints an example command that could be used to install the desktop version of Qt.

Todo:
* [x] Fix the Updater for Windows. Currently, for Android installations on Windows, the Updater assumes that the desktop version of Qt is `mingw81_64`, regardless of what is already installed, what was installed by `--autodesktop`, and what is actually available. This is a serious bug: if aqt automatically installs mingw73_64, but patches the android files to look for some other version, the patch won't help anyone.
* [x] Improve test coverage.
* [x] Add 1-2 jobs to the Azure Pipelines that use this flag. I don't want to add too much here, because the `--autodesktop` flag does not allow the user to filter out archives; this means that every new job will download hundreds of MBs and take much longer to run than most of the other jobs.

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -35,10 +35,10 @@ from logging import getLogger
 from logging.handlers import QueueHandler
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 import aqt
-from aqt.archives import QtArchives, QtPackage, SrcDocExamplesArchives, ToolArchives
+from aqt.archives import QtArchives, QtPackage, SrcDocExamplesArchives, TargetConfig, ToolArchives
 from aqt.exceptions import (
     AqtException,
     ArchiveChecksumError,
@@ -299,7 +299,23 @@ class Cli:
             if modules is not None and archives is not None:
                 archives.append(modules)
         nopatch = args.noarchives or (archives is not None and "qtbase" not in archives)  # type: bool
-        warn_on_missing_desktop_qt: bool = not args.autodesktop
+        should_autoinstall: bool = args.autodesktop
+        _version = Version(qt_version)
+        base_path = Path(base_dir)
+
+        expect_desktop_archdir, autodesk_arch = self._get_autodesktop_dir_and_arch(
+            should_autoinstall, os_name, target, base_path, _version
+        )
+
+        auto_desktop_archives: List[QtPackage] = (
+            retry_on_bad_connection(
+                lambda base_url: QtArchives(os_name, "desktop", qt_version, autodesk_arch, base=base_url, timeout=timeout),
+                base,
+            ).archives
+            if autodesk_arch is not None
+            else []
+        )
+
         if not self._check_qt_arg_versions(qt_version):
             self.logger.warning("Specified Qt version is unknown: {}.".format(qt_version))
         if not self._check_qt_arg_combination(qt_version, os_name, target, arch):
@@ -310,7 +326,7 @@ class Cli:
         if not all_extra and not self._check_modules_arg(qt_version, modules):
             self.logger.warning("Some of specified modules are unknown.")
 
-        qt_archives = retry_on_bad_connection(
+        qt_archives: QtArchives = retry_on_bad_connection(
             lambda base_url: QtArchives(
                 os_name,
                 target,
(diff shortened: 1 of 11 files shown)


aqt/installer.py
```diff
@@ -632,6 +640,43 @@ def _warn_on_deprecated_parameter(self, parameter_name: str, value: str):
...
+            f"is also installed."
+        )
+        if should_warn:
+            self.logger.warning(
+                f"{msg_prefix} You can install it with the following command:\n"
+                f"          `aqt install-qt {host} desktop {version} {missing_desktop_arch}`"
+            )
+        else:
+            self.logger.info(f"{msg_prefix} Now installing Qt: desktop {version} {missing_desktop_arch}")
+            self.run(["install-qt", host, "desktop", format(version), missing_desktop_arch])
```
ddalcino, line 678: I am thinking about changing this to:
```suggestion
            self.run(["install-qt", host, "desktop", format(version), missing_desktop_arch, "--archives", "qtbase", "qtdeclarative"])
```
... to save on network traffic, because that's what our current CI script does.

I'm not sure if this is the best idea though. I'm not sure which archives are actually required for android/ios, and I don't want our users to be left with an installation that doesn't work.

aqt/metadata.py
```diff
@@ -351,6 +353,65 @@ def possible_extensions_for_arch(arch: str) -> List[str]:
...
+
+        def contains_qmake_exe(arch_path: Path) -> bool:
+            return (arch_path / "bin/qmake.exe").is_file()
+
+        paths = [d for d in installed_qt_version_dir.glob("mingw*")]
+        directories = list(filter(contains_qmake_exe, paths))
+        arches = [d.name for d in directories]
+        selected = QtRepoProperty.select_default_mingw(arches, is_dir=True)
+        return installed_qt_version_dir / selected if selected else None
+
```
ddalcino, line 414: `QtRepoProperty` is now 100 lines, and it's all static methods. It doesn't really have anything to do with the MetadataFactory. Maybe it should be a module instead of a class, and exist in its own file?

ddalcino:
I'm not sure what's going on with the CI run here. The `tox` run is failing because of a problem loading `isort`; it's being tracked in several places, including here: https://github.com/gforcada/flake8-isort/issues/115 There's a workaround people are talking about, and I think it involves pinning the versions of `flake8` and `flake8-isort`. I am unable to get it to work though.

aqt/installer.py
ddalcino: The Azure Pipelines jobs don't appear to run any slower than they did before.

ddalcino:
I think this PR is done for now, except for a couple of things:

1. It should probably be squashed into fewer commits
2. I’m not really happy about the way QtRepoProperty is turning out. It definitely needs its own file, but right now it cannot be moved without breaking out Version and a couple other things without creating dependency cycles. This would split metadata.py into at least 4 different files.
3. There’s some question about whether—autodesktop should be opt-in or opt-out. Ideally, it should be opt-out, but that cannot be done without breaking current interface

miurahr:
I'd like to publish this improvements as v3.0.0beta/rc versions and ask users to test their environments.
It is better to recreate  commit history for future maintenance before merge.

miurahr:
I'd like to merge here to master. @ddalcino could you re-create commit history and fix merge conflict?

miurahr:
I think it worth squashing commits to merge.

ddalcino:
> I think it worth squashing commits to merge.

Ok. I've re-ordered and squashed about half of these commits, and I think it's worthwhile to preserve the remaining history. I'm willing to squash it further if you like.

We should probably discuss the default behavior of the `--autodesktop` flag. Right now, it is opt-in, because I was expecting this PR to go into aqt v2.3 and not 3.0. There was some discussion of this issue here: https://github.com/miurahr/aqtinstall/issues/528#issuecomment-1193364530.

Since we are now talking about aqt v3.0, do you think we should change the flag to `--no-autodesktop`, and make this an opt-out feature? There are probably other ways to modify the default behavior, such as `settings.ini` entries.

miurahr:
I'd like to merge it as-is.

PR #541, opened 2022-07-24
Update Mac CI pipelines to use MacOS-12

ddalcino:
The MacOS-10.15 CI runners are now deprecated, and will begin brownouts in August. See discussion here: https://github.com/actions/virtual-environments/issues/5583

~MacOS-11 is also available, in case there are any issues with 12.~ There are issues with 12 so I went to 11.

**Update:** Something's wrong with MacOS-11 too: https://dev.azure.com/miurahr/github/_build/results?buildId=5481&view=logs&j=bae13455-f657-522d-e2b4-64495419fb63&t=5c902093-735e-5ee6-b7c8-ba98a93e04e2&l=58

**Update [2021-09-17]:** I have come to the conclusion that it makes no sense to continue to test using Qt < 6.2.4 with ios, because it is not realistic that it will work. The XCode toolchain has changed significantly in the last few years, breaking compatibility with several things that older versions of Qt relied on in the past. Every version of Qt greater than or equal to 6.2.4 works without any trouble in CI on MacOS 12.

Technically, you can modify certain files in Qt versions 6.0-6.2.3 and end up with a successful ios build (install Python 2, change a shebang line in `ios/mkspecs/features/uikit/devices.py`, and modify xml entries in `ios/mkspecs/macx-xcode/WorkspaceSettings.xcsettings`), but I strongly oppose advising our users to do these things, rather than simply using a version that works properly to begin with.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,10 +157,10 @@ to make it working well with installed directory.
 
 .. warning::
    If you are using aqtinstall to install the ios version of Qt, please be aware that
-   there are compatibility issues between XCode 13+ and versions of Qt less than 6.2.0.
+   there are compatibility issues between XCode 13+ and versions of Qt less than 6.2.4.
    You may use aqtinstall to install older versions of Qt for ios, but the developers of
    aqtinstall cannot guarantee that older versions will work on the most recent versions of MacOS.
-   Aqtinstall is tested for ios on MacOS 12 with Qt 6.2.0 and greater.
+   Aqtinstall is tested for ios on MacOS 12 with Qt 6.2.4 and greater.
    All earlier versions of Qt are expected not to function.
 
 Testimonies
--- a/README.rst
+++ b/README.rst
@@ -155,6 +155,14 @@ to make it working well with installed directory.
 .. note::
    It is your own task to set some environment variables to fit your platform, such as PATH, QT_PLUGIN_PATH, QML_IMPORT_PATH, and QML2_IMPORT_PATH. aqtinstall will never do it for you, in order not to break the installation of multiple versions.
 
+.. warning::
+   If you are using aqtinstall to install the ios version of Qt, please be aware that
+   there are compatibility issues between XCode 13+ and versions of Qt less than 6.2.0.
+   You may use aqtinstall to install older versions of Qt for ios, but the developers of
+   aqtinstall cannot guarantee that older versions will work on the most recent versions of MacOS.
+   Aqtinstall is tested for ios on MacOS 12 with Qt 6.2.0 and greater.
+   All earlier versions of Qt are expected not to function.
+
 Testimonies
 -----------
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,7 +32,7 @@ jobs:
 - job: Mac
   dependsOn: MatricesGenerator
   pool:
-    vmImage: 'macOS-10.15'
+    vmImage: 'macOS-12'
   strategy:
     matrix: $[ dependencies.MatricesGenerator.outputs['mtrx.mac'] ]
   steps:
--- a/ci/generate_azure_pipelines_matrices.py
+++ b/ci/generate_azure_pipelines_matrices.py
@@ -270,7 +270,7 @@ mac_build_jobs.append(
 # mobile SDK
 mac_build_jobs.extend(
     [
-        BuildJob("install-qt", "5.15.2", "mac", "ios", "ios", "ios", is_autodesktop=True),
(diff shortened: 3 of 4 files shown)


ddalcino:
Maybe it's too early to go up to MacOS-12 yet: [this ios build](https://dev.azure.com/miurahr/github/_build/results?buildId=5480&view=logs&jobId=bae13455-f657-522d-e2b4-64495419fb63&j=1ea727ed-2050-5877-9d96-adddb35f7943&t=b67eabc5-7e7e-54c5-0f88-bc35772ccffe) failed, because there's a file at `Qt/6.2.2/ios/mkspecs/features/uikit/devices.py` that specifies that it must be run with `/usr/bin/python` in the shebang. MacOS-12 removed `/usr/bin/python` in 12.3, because it was a version of Python 2. Python 3.8 is still at `/usr/bin/python3`.

I am not sure if we can (or should) get around this by adding a symlink at /usr/bin/python to python3. It's probably not a good idea, and not something we should expect our users to do. I think this is probably something the Qt devs should fix.

ddalcino:
Ok, I don't know why the ios builds are using a "legacy build system" in CI, but until we figure out how to use the "new build system", our ios builds are going to fail in CI. There are docs on apple.com that describe the fix using a GUI interface here: https://help.apple.com/xcode/mac/current/#/dev396bc94c7. I do not see any way to choose the build system in a headless environment. Maybe this can be done in Applescript?

It's hard for me to see this issue as anything other than a bug in the MacOS-11 environment.

ddalcino:
I took another look at `Qt/*/ios/mkspecs/features/uikit/devices.py`, and here's what I've found so far:
* The shebang line in this file, for every version of Qt that I have looked at starting at 6.3.0, requests `/usr/bin/env python3`. Additionally, the 7z archive filenames indicate that they have been built for MacOS 11 or 12.
* For Qt 6.2.4, and several earlier versions that I checked, the shebang line specifies `/usr/bin/python`, which is generally Python 2 on MacOS. The 7z archive filenames indicate MacOS 10.15 or earlier.
* For some reason, despite requesting `/usr/bin/python` in the `devices.py` file, the Qt 6.2.2 job on MacOS 11 and 12 passed, while the Qt 5.15.2 job did not. I cannot explain this.

I am getting the impression that Python 2 is necessary to complete the older ios builds successfully. There are ways to do this on modern MacOS; we could try using pyenv.

However, keep in mind that every version of Qt 6.1 and below is now past end-of-life. Qt 5.15.* LTS is not past end-of-life, but the newest versions aren't available via aqtinstall (they are paid only). We should seriously consider dropping support for the older ios versions. It's probably not a good idea to install Python 2 in CI, or try to convince the system that Python 3 will do the same job as Python 2, as suggested earlier.

ddalcino:
Ok, I've gotten a little further, and here's what I've found: 

I was able to install Python 2, and convince the Python 2 files in Qt to use the right version of Python. This helped, but it was not sufficient to fix the failing builds.

The Qt 5.15.2 and 6.0.4 binaries available on download.qt.io are built for MacOS 10.14, which was released four years ago. `qmake` from Qt 5.15.2 and 6.0.4 will build a Makefile that relies on the 'legacy build mode' of the XCode toolchain, for every version of XCode since version 10 (the version of XCode that works with MacOS 10.14). 'Legacy build mode' was the default build mode in MacOS 10.14. It is not the default in MacOS 11 or 12. You can turn it on with a command line flag, but I'm not sure how to do that in CI. If we want to continue testing Qt 5.15.2 and 6.0.4 in CI, we will need to find a way to turn this flag on.

Unfortunately, legacy build mode will be removed in a future version of MacOS. Any effort we put into this will be undone as soon as Apple removes legacy mode.

The Qt 6.1.4 build is a little different: The Qt 6.1.4 binary on download.qt.io was built for MacOS 11, and it does not require legacy build mode. The Accelbubble example mostly builds OK, but it fails when it tries to link to the `-lqtsensors_ios` library. There's a `qtsensors` module for every version of Qt 6.2 and greater, but I cannot find any evidence of a such a library anywhere in Qt 6.1.4. I think that if we want to test Qt 6.1.4 for ios, we should use an example that does not use `qtsensors`.

I think that at this point, we should only test ios builds for Qt 6.2 and up. It's only going to get harder to support older versions Qt for ios, and I don't think the effort is worth it. We should include warnings in our documentation about these issues, and state very clearly that we cannot provide support when users have trouble with Qt for ios 6.1.4 and below.

miurahr:
I'd like to squash and merge. Thank you for hard work!

PR #545, opened 2022-08-07
Make MetadataFactory respect baseurl set in arguments

ddalcino:
This fixes an issue that I had noticed and promptly forgot about: https://github.com/miurahr/aqtinstall/issues/519#issuecomment-1109875714

> After a brief review of the code, it looks like the list-* commands are not capable of setting the base url via CLI argument. This has consequences: at least one install-tool command allows setting base url via CLI argument, and calls MetadataFactory functions, which do not respect the chosen base url. This will need to be fixed.

Before this change, any time `aqt install-qt` or `aqt install-tool` used `MetadataFactory` to resolve a Qt version spec or fetch a list of tools, it would ignore whatever mirror base url the user sets with the `--base` flag and use `Settings.baseurl` instead. This PR fixes that.

Note: There's a big ugly addition to `tests/test_install.py`. I'm sorry about that. I can refactor the test to reuse the existing `test_install()` function, but that function is already big and complicated, and I do not want to make it worse.

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -197,7 +197,9 @@ class Cli:
         return all([m in available for m in modules])
 
     @staticmethod
-    def _determine_qt_version(qt_version_or_spec: str, host: str, target: str, arch: str) -> Version:
+    def _determine_qt_version(
+        qt_version_or_spec: str, host: str, target: str, arch: str, base_url: str = Settings.baseurl
+    ) -> Version:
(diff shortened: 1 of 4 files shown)


miurahr:
LTGM.
 It is ok for the addition on test,  because an effort to reduce common code for test code does not produce many value and it is more important to show what the tests want to see  behaviors  or results in simple, stupid way, that is easy to understand.

miurahr:
I'd like to merge #545 and publish it as v2.2.1.

PR #557, opened 2022-08-15
Fix windows standalone binary build

nikitalita:
* update build_standalone.bat to rename pyproject.toml to pyproject.toml.bak to ensure pip installs from setup.cfg
* update dependencies in `setup.cfg`


miurahr:
Thank you for workaround.

PR #571, opened 2022-09-19
Clean duplication in setup.cfg

miurahr:
New pip versions and standalone binary build supports `pyproject.toml`.
We can drop setup.py and setup.cfg, at least we can remove its duplicated configurations.


miurahr:
#598 does clean-up.

PR #574, opened 2022-09-23
Update documents for release 3.0.0

miurahr:
- Add --autodesktop example on README
- Update patch version in examples
- Add missing changes


ddalcino:
Is https://aqtinstall.readthedocs.io/en/stable/ up to date? From the changelog entries, it looks like it has not been updated since aqt v2.1.0 was released in April.

PR #581, opened 2022-09-30
Actions: fix upload artifact script

miurahr:


miurahr:
Binary should be `aqt_x64.exe` not `aqt-x64.exe` for consistency.

PR #598, opened 2022-10-28
Standalone binary build with PyInstaller

miurahr:
- Call PyInstaller directly
- Move build_standalone.py into tools folder
- Remove duplicated configurations from setup.cfg


miurahr:
I made standalone_build script simpler.

@nikitalita could you check a new standalone binary work for you?

nikitalita:
It works, but only after I install the venv manually. I'm kind of surprised that the CI works with bash though; the default Windows runner uses Git Bash, and I've found that using the default Windows python with it will sometimes cause it to behave very oddly. I'd recommend setting up a powershell script for the venv setup steps, as powershell works on all platforms.

miurahr:
@nikitalita Even build on powershell and remove venv after build, it passed.
Could you share reproduce conditions or logs?

miurahr:
@nikitalita  You can download artifact which is built on CI https://github.com/miurahr/aqtinstall/actions/runs/3349589622#artifacts

miurahr:
I've squashed commits.

tools/launch_aqt.py
```diff
@@ -0,0 +1,8 @@
+#!python.exe
+import aqt
+import re
+import sys
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
```
miurahr, line 7: Is it better to be just `sys.argv[0] = 'aqt'` ?

tools/build_standalone.py
```diff
@@ -0,0 +1,33 @@
...
+        adddata_arg = "{src:s};aqt"
+    else:
+        adddata_arg = "{src:s}:aqt"
+    for data in ["aqt/logging.ini", "aqt/settings.ini", "aqt/combinations.json"]:
+        args.append('--add-data')
+        args.append(adddata_arg.format(src=data))
+    args.append(os.path.join(tools_dir, "launch_aqt.py"))
+
+    # launch PyInstaller
+    PyInstaller.__main__.run(args)
```
miurahr, line 33: Here is a way to call PyInstaller from python code.
https://pyinstaller.org/en/stable/usage.html#running-pyinstaller-from-python-code

It is better than `gravitybee` that call `pyinstaller` using `subprocess`.
Now the process is speedy and safe.

nikitalita:
> @nikitalita Even when build on powershell and remove venv after built, it passed on CI. Could you share reproduce conditions or logs?

It works fine. I'm just saying that it MIGHT fail in the future if we keep using `bash`.

tools/launch_aqt.py
miurahr: It is ok for now.

miurahr:
Thank you for feedback. I'd like to merge.

PR #600, opened 2022-10-29
Actions: ci test with v2 command arguments

miurahr:
use install-qt instead of install


miurahr:
V2 commands are tested on Azure CI. So GH actions test v1 one.

PR #608, opened 2022-11-05
improve link to report issue

EtlamGit:
When we print the link with trailing dot, some automatic log-formaters will add that dot to the link itself.
E.g. when you use Github Actions this link will show up with extra dot and therefore does not work.
By removing the dot, the link is functional also in these cases.


miurahr:
Thank you @EtlamGit . Good catch!

EtlamGit:
It was an easy catch, as I encountered that problem myself...

PR #612, opened 2022-11-15
Fix expected error message in test cases

ddalcino:
This fixes two failing tests, as reported here: https://github.com/miurahr/aqtinstall/commit/924e4b087bf509ab62c7d6588dc358d0e31de89e#comments.


miurahr:
Thank you for the help! @ddalcino

PR #619, opened 2022-11-20
Add snyk security scanner

miurahr:


miurahr:
It fails because snyk scanner don't support PEP517 project yet.
https://github.com/snyk/snyk-python-plugin/issues/189

PR #631, opened 2023-01-01
Fix failing disabled test `test_cli_unknown_version`

ddalcino:
It looks like this test was broken when the logging format changed, but since it is marked as `remote_data`, it never runs in CI. This change should allow the test to pass.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -37,9 +37,9 @@ def test_cli_unknown_version(capsys):
     """
 
     matcher = re.compile(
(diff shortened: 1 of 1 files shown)


ddalcino:
Not sure what's wrong with the CI run here. I can reproduce the issue locally, in this branch and in the master branch. I'm certain that this error has nothing to do with the changes in this PR, but I have not figured out why it's happening.

ddalcino:
Superseded by #641

PR #634, opened 2023-01-08
Update tox.ini config

miurahr:
tox 4 accept comma separate value or array for pass_env, and raise error for space separate values

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,9 +177,9 @@ isolated_build = True
 
 [testenv]
 passenv =
-    GITHUB_*
-    APPVEYOR APPVEYOR_*
-    TRAVIS TRAVIS_*
+    GITHUB*
+    APPVEYOR*
+    TRAVIS*
     COVERALLS_*
     PYTEST_ADDOPTS
 extras = test, check, docs
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ aqt = "aqt.__main__:main"
 test = [
     "pytest>=6.0",
     "pytest-pep8",
-    "pytest-remotedata",
+    "pytest-remotedata<0.4.0",
     "pytest-cov",
     "pytest-socket",
     "pytest-leaks",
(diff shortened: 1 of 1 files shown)


ddalcino:
Looks like a big improvement. However, I think it's failing now due to a separate issue:

https://github.com/miurahr/aqtinstall/actions/runs/3867790752/jobs/6592802611#step:7:261
```
argparse.ArgumentError: argument -R/--remote-data: conflicting option string: -R
```

I think this is a CLI conflict between `pytest-leaks` and `pytest-remotedata`. Both of them define `-R` as a command line option. I think `pytest` is trying to build a command line interface that includes both options. I cannot run pytest locally with both of them installed, but I can run it after I uninstall `pytest-leaks`.

miurahr:
I'd like to remove `pytest-leaks`

ddalcino:
> I'd like to remove `pytest-leaks`

It probably makes sense to do this temporarily, but in the long term, it would be nice to add it back in (if possible).

It would be really nice to know what happened to cause this. We were using pytest-leaks along with pytest-remotedata for a long time without any problems, weren’t we? Is this a new change in pytest, or in one of the plugins? I would file an issue with the appropriate project, but I don’t know which one is causing this.

It would be nice to be able to turn off single-letter options like ‘-R’ that are likely to cause conflicts.

pyproject.toml
```diff
@@ -216,7 +215,6 @@ commands =
     python3.9-dbg -m pytest -v --no-cov -R : -k "test_install"
 deps =
     pytest
-    pytest-leaks
     pytest-remotedata
```
ddalcino, line 218: I noticed that in the last working CI runs, `pytest-remotedata` was at version 0.3.3. When I tried installing `pytest-leaks` and `pytest-remotedata==0.3.3`, pytest would run properly again!

Try this:
```suggestion
    pytest-leaks
    pytest-remotedata<0.4.0
```

ddalcino (approved):
LGTM. I don't see any reason not to merge this. If we don't merge this, there will be no way for tests to pass in other PRs.

PR #641, opened 2023-02-03
Pin EMSDK version for build where useful

iakov:
Also fix test that fails at the moment

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -37,7 +37,7 @@ def test_cli_unknown_version(capsys):
     """
 
     matcher = re.compile(
-        r"^INFO    : aqtinstall\(aqt\) v.* on Python 3.*\n"
+        r"[^\n]*aqtinstall\(aqt\) v.* on Python 3.*\n"
         r".*Specified Qt version is unknown: " + re.escape(wrong_version) + r"\.\n"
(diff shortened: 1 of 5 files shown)


iakov:
This small PR just helps to make all checks happy

tests/test_connection.py
```diff
@@ -37,7 +37,7 @@ def test_cli_unknown_version(capsys):
     """
 
     matcher = re.compile(
-        r"^aqtinstall\(aqt\) v.* on Python 3.*\n"
+        r"^INFO    : aqtinstall\(aqt\) v.* on Python 3.*\n"
```
ddalcino, line 40:
```suggestion
        r"[^\n]*aqtinstall\(aqt\) v.* on Python 3.*\n"
```
I think this change would allow the `INFO:` prefix, but not require it. This should make the test more resilient to differences in dev environment and potential changes to the format of the prefix.

ddalcino:
Thanks for fixing this! Looks like the WASM builds are fixed now.

It looks like CI is failing right now because `black` would make changes to a couple of test files (test_helper.py and test_cli.py) that you have not changed. I don't understand why that is; I thought they were working before. Maybe `black` was updated recently?

IMHO this is worth merging as is, but I would not mind if you wanted to run `black` on the `test/` directory.

ddalcino (approved):
Nice work! Thank you!

PR #649, opened 2023-02-11
Pin checkout at `v3` in all workflows

ddalcino:
This fixes some CI warnings about using obsolete versions of NodeJS. NodeJS 12 (from `actions/checkout@2`) is at EOL, and CI warnings recommend using NodeJS 16. The docs for actions/checkout recommend pinning at v3.

Some of these workflows were using `checkout@master`, and were thereby already using NodeJS 16. I'm not sure that it's necessary to use v3 for these instead of master. However, the docs suggest using v3. If at some point the authors decide to rename the `master` branch to `main`, then we may be glad we made this change.

### DIFF
--- a/.github/workflows/check-combinations.yml
+++ b/.github/workflows/check-combinations.yml
@@ -14,7 +14,7 @@ jobs:
   check_combinations:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


miurahr:
good catch, thank you!

Issue #578, opened 2022-09-29
Installation failed even when new version listed on list-qt (just released 6.4.0 today)


DioEgizio:
**Describe the bug**
can't download 6.4.0 even if it's out and can be downloaded with the official installer

**To Reproduce**
1. run `aqt` with this command: python3 -m aqt install-qt mac desktop 6.4.0
2. See error 

ERROR   : Failed to locate XML data for Qt version '6.4.0'.
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt mac desktop' to show versions available.

**Expected behavior**
it working 

**Desktop (please complete the following information):**
 - OS: macOS
 - `aqt` version v2.2.3

**Additional context**
it made me update the mantainance tool trying to update with the official installer, perhaps smth changed there?

ddalcino:
I will have a look at this when I get home from work, but you are right that this is a problem. `aqt list-qt` should not be allowed to list items that you cannot successfully install.

I have noticed that there have been problems recently with installing Qt or tools less than a day after they have been released, because there’s a delay between the time the server uploads files and when the sha256 hashes for those files are available. Right now, `aqt install-*` won’t install anything without the hash files for security reasons. `aqt list-*` does not check for hash files at all, but it probably should. See #577, #521

The most recent update for Qt 6.4.0 is dated about 2 hours before this issue was created. By the time I get off work and take a look at this, the hash files might be up on the server and you’ll be able to install this properly again. I don’t know how long the delays are, but I know they are less than 24 hours. They could be less than 12.

miurahr:
Qt download site may create hashes with hashing command, not on-the-fly.

MirrorBrain, which Qt project uses, manual explain how to create hash, to advice to use cron job.
https://mirrorbrain-docs.readthedocs.io/en/latest/installation/initial_config.html#creating-hashes

> Add the hashing command to /etc/crontab to be run every few hours. Alternatively, run it after changes in the file tree happen, coupled to some trigger etc.

A job calculating hash requires reading all the binaries, it will take some durations.

miurahr:
I push a pin the issue because many other users may want to use aqtinstall just after new Qt releases.
It will become FAQ.

ddalcino:
Sorry it took me so long to get to this, but I have confirmed that aqt v2.2.3 can now install Qt 6.4.0 properly, after the hashes were uploaded.

<details>
<summary>$ aqt install-qt mac desktop 6.4.0</summary>
<pre>
[...]

PR #677, opened 2023-05-16
Add support for standalone sdktool installation

luk1337:
This allows us to install sdktool using the following command:
aqt install-tool windows desktop sdktool

### DIFF
--- a/aqt/combinations.json
+++ b/aqt/combinations.json
@@ -547,6 +547,12 @@
     ],
     "tools": [
       {
+        "arch": "qt.tools.qtcreator",
+        "os_name": "linux",
+        "target": "desktop",
+        "tool_name": "sdktool"
+      },
+      {
         "arch": "qt.tools.cmake",
         "os_name": "linux",
         "target": "desktop",
@@ -685,6 +691,12 @@
         "tool_name": "tools_telemetry_gui"
       },
       {
+        "arch": "qt.tools.qtcreator",
+        "os_name": "mac",
+        "target": "desktop",
+        "tool_name": "sdktool"
+      },
+      {
         "arch": "qt.tools.cmake",
         "os_name": "mac",
         "target": "desktop",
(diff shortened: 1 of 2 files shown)


miurahr:
thank you for the proposal! 
when check combinations.json, there are only entries for sdktools.doc not sdktool
https://github.com/miurahr/aqtinstall/blob/master/aqt/combinations.json#L814

Could you also improve CI tool to update the data?
https://github.com/miurahr/aqtinstall/blob/master/ci/generate_combinations.py#L54

`iter_tool_variants` method seems return no "sdktool" variant.

A tool is  called by GitHub Actions
https://github.com/miurahr/aqtinstall/blob/master/.github/workflows/check-combinations.yml

luk1337:
It seems like `ci/generate_combinations.py` is already aware of `sdktool` without doing anything. However, it was failing at `tools_qtdesignstudio_generation2`, which I had to temporarily blacklist when running it.

miurahr:
> It seems like `ci/generate_combinations.py` is already aware of `sdktool` without doing anything. However, it was failing at `tools_qtdesignstudio_generation2`, which I had to temporarily blacklist when running it.

Thank you for the trial. You are welcome to change to add it blacklist in another PR.

luk1337:
> > It seems like `ci/generate_combinations.py` is already aware of `sdktool` without doing anything. However, it was failing at `tools_qtdesignstudio_generation2`, which I had to temporarily blacklist when running it.
> 
> Thank you for the trial. You are welcome to change to add it blacklist in another PR.

Maybe `tools_qtdesignstudio_generation2` will eventually get fixed, it just seems weird that file exists but .sha256 doesn't.
* https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_qtdesignstudio_generation2/Updates.xml
* https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_qtdesignstudio_generation2/Updates.xml.sha256

ddalcino:
> Maybe `tools_qtdesignstudio_generation2` will eventually get fixed, it just seems weird that file exists but .sha256 doesn't work.
> 
>     * https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_qtdesignstudio_generation2/Updates.xml
> 
>     * https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/tools_qtdesignstudio_generation2/Updates.xml.sha256

Unfortunately, missing `.sha256` files are common; it's an existing issue that does not have an easy solution (see #578). However, they are typically missing for less than 24 hours after new files are uploaded. If you check the sha256 link again, you'll find that it's working now.

PR #680, opened 2023-05-23
Fixed command to check tools_mingw90 version

bjorn:
Document command gives:
```
aqt: error: unrecognized arguments: --long-modules tools_mingw90
```


ddalcino:
Thanks for the PR, it’s a nice catch!

This seems to be a quirk of argparse; it allows you to write flags and options either before all the positional arguments or after them. I don’t think I (we?) have been particularly consistent with the way we write these in the docs; it might be worthwhile to audit all the example commands for problems like this. I bet there are more!

ddalcino (changes requested):
Please use the long form of the `-l` flag.

bjorn:
> Please use the long form of the `-l` flag.

Makes sense. Feel free to squash when merging. :-)

Issue #687, opened 2023-06-02
"Issue with aqtinstall Command: Unresponsive Download Link"


LowSugarCoke:
**Describe the bug**
I am encountering an issue while using the aqtinstall command as mentioned below. Specifically, the download link seems to be unresponsive or malfunctioning.
```
python -m aqt install-qt --outputdir C:\\test windows desktop 5.15.2 win32_msvc2019 -m qtquicktimeline
```

**To Reproduce**
1. install `aqt` using this command
```
python -m pip install aqtinstall
```
2. run `aqt` with this command
```
python -m aqt install-qt --outputdir C:\\test windows desktop 5.15.2 win32_msvc2019 -m qtquicktimeline
```
3. See error
```
C:\Users\leeja\Desktop\MonitorControlPortalMaster\AzurePipeline\template>python -m aqt install-qt --outputdir C:\\test windows desktop 5.15.2 win32_msvc2019 -m qtquicktimeline
INFO    : aqtinstall(aqt) v3.1.6 on Python 3.10.4 [CPython MSC v.1929 64 bit (AMD64)]
INFO    : Downloading qtconnectivity...
INFO    : Downloading qtgraphicaleffects...
INFO    : Downloading qtquicktimeline...
INFO    : Downloading qtmultimedia...
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Downloading qtquicktimeline: attempt #2...
INFO    : Downloading qtgraphicaleffects: attempt #2...
INFO    : Downloading qtmultimedia: attempt #2...
INFO    : Downloading qtconnectivity: attempt #2...
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Redirected: mirrors.sjtug.sjtu.edu.cn
INFO    : Downloading qtquicktimeline: attempt #3...
INFO    : Downloading qtgraphicaleffects: attempt #3...
INFO    : Downloading qtmultimedia: attempt #3...
INFO    : Downloading qtconnectivity: attempt #3...
[...]

PR #688, opened 2023-06-02
fix: remove a mirror that can be error

miurahr:
- remove https://mirrors.sjtug.sjtu.edu.cn/qt/ from fallbacks
- Adjust from report #687


pzhlkj6612:
Maybe we can make aqt switch between different mirror servers during the download?

miurahr:
> Maybe we can make aqt switch between different mirror servers during the download?

We maybe able to add a mirror URL as black list `in live` when some download filed and re-try other mirror, so we can change a mirror selection during download.

PR #689, opened 2023-06-18
Docs: might be unhappy with the default blacklist settings

pzhlkj6612:
Hello!

---

I tried "```sphinx-intl update -l ja```" after editing the docs, but git-diff told me that the changes are a lot and not just the lines I changed, so I didn't commit them in this PR.


miurahr:
Thank you!

Issue #696, opened 2023-07-25
aqt install fail with a 7zip error


yannou38:
We have a kind of weird CI using github action executing on runners located on a laptop, or on a azure VM. The runners share some installed programs, such as QT.

We have a github action to install QT on the laptop or the azure VM. The installation works properly on the laptop, but fail on the VM with the following log:

```
Notice:  A new release of pip available: 22.3.1 -> 23.2.1
Notice:  To update, run: python.exe -m pip install --upgrade pip
INFO    : aqtinstall(aqt) v3.1.6 on Python 3.11.1 [CPython MSC v.1934 64 bit (AMD64)]
ERROR   : Specified 7zip command executable does not exist: '7z'
```

Could you advise ? this error doesn't seem common from what i could search, but happen every time the github action execute on one of the azure VM runner.

The command executed is aqt install-qt windows desktop 6.3.0 win64_msvc2019_64 --outputdir foo

yannou38:
Note: here is the install log for aqt on the azure VM:

`
Notice:  A new release of pip available: 22.3.1 -> 23.2.1
Notice:  To update, run: python.exe -m pip install --upgrade pip
Collecting aqtinstall
  Using cached aqtinstall-3.1.6-py3-none-any.whl (59 kB)
Collecting bs4
  Using cached bs4-0.0.1.tar.gz (1.1 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting defusedxml
  Using cached defusedxml-0.7.1-py2.py3-none-any.whl (25 kB)
Collecting humanize
  Using cached humanize-4.7.0-py3-none-any.whl (113 kB)
Collecting patch>=1.16
  Using cached patch-1.16.zip (12 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting py7zr>=0.20.2
  Using cached py7zr-0.20.5-py3-none-any.whl (66 kB)
Collecting requests>2.20.0
  Using cached requests-2.31.0-py3-none-any.whl (62 kB)
Collecting semantic-version
  Using cached semantic_version-2.10.0-py2.py3-none-any.whl (15 kB)
Collecting texttable
  Using cached texttable-1.6.7-py2.py3-none-any.whl (10 kB)
Collecting pycryptodomex>=3.6.6
  Using cached pycryptodomex-3.18.0-cp35-abi3-win_amd64.whl (1.7 MB)
[...]

PR #705, opened 2023-08-27
Log a warning when aqtinstall falls back to an external 7z extraction tool

ddalcino:
Inspired by https://github.com/miurahr/aqtinstall/issues/696#issuecomment-1651026317. That comment suggested a place to add a log message, but I was unable to log the message there without disregarding the user’s selected log settings, which are not available until the `Cli` object is built.

When `aqtinstall` fails to load the `py7zr` module, it falls back to whatever 7z extraction tool the user has specified in the `settings.ini` file, without notifying the user. This PR adds the following log message when that happens:

```
WARNING: The py7zr module failed to load. Falling back to '7z' for .7z extraction.
WARNING: You can use the  '--external | -E' flags to select your own extraction tool.
```

This PR also does a little refactoring of the `Cli._set_sevenzip` method to improve readability.

**Side effect**
There was an existing bug here, where the `install-qt` subcommand (but none of the other commands) would ignore the `7zcmd` parameter in the `settings.ini` file, and use the hardcoded `"7z"` instead. I'm pretty sure this was a bug; if not I can change it back.

https://github.com/miurahr/aqtinstall/blob/72da54f4b0156a1bbd0aea09c621992817f963f8/aqt/installer.py#L361-L363

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -233,10 +233,11 @@ class Cli:
             "This may not install properly, but we will try our best."
         )
 
-    def _set_sevenzip(self, external: Optional[str], fallback: str, is_p7zr_missing: bool) -> Optional[str]:
+    def _set_sevenzip(self, external: Optional[str]) -> Optional[str]:
         sevenzip = external
+        fallback = Settings.zipcmd
         if not sevenzip:
-            if is_p7zr_missing:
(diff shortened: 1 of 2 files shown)


miurahr:
Thank you for the improvement.

PR #707, opened 2023-08-28
Use 'tar' filter when extracting tarfiles

### LINKED ISSUE #700: potential security issue in tar extraction
Synk warned

> Calling extractall to extract all files from a tar file without sanitization. This may result files outside destination directory to be overwritten, resulting in an arbitrary file write.
https://github.com/miurahr/aqtinstall/blob/ebc8175deb13eb24bf8acf57533038ec21dd0c77/aqt/installer.py#L1229

python `tarfile` provide filter function for sanitization. it can be

```
  tar_archive.extractall(path=base_dir, filter='tar') 
```

ddalcino:
Fix #700.

This is to address a Snyk warning. The new code will only run on Python 3.11.4 and above.

Tested locally with Python 3.11.2 and 3.11.5. Although we can't see it in the code coverage with a single version of Python, I can confirm personally that both arms of the conditional are being entered.

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -1226,7 +1226,12 @@ def installer(
 
     if tarfile.is_tarfile(archive):
         with tarfile.open(archive) as tar_archive:
-            tar_archive.extractall(path=base_dir)
+            if hasattr(tarfile, "data_filter"):
+                tar_archive.extractall(filter="tar", path=base_dir)
+            else:
+                # remove this when the minimum Python version is 3.12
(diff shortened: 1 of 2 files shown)


ddalcino:
Oops; looks like this change caused this failure on Windows: https://github.com/miurahr/aqtinstall/actions/runs/5997958636/job/16265345067

On that CI runner, it looks like Python 3.10 is installed, and the test is not expecting to see the warning `Extracting may be unsafe; consider updating Python to 3.11.4 or greater`. Let me fix that...

Issue #731, opened 2023-11-10
list_* commands ignore base url setting


comargo:
https://github.com/miurahr/aqtinstall/blob/d47adb88b0ae78bfc091298d37224ee0106e4fd5/aqt/metadata.py#L524C31-L524C31

Since Python assigns default values as soon as it imports corresponding module (executes `def func():` line), and Settings file has not been read yet, the MetadataFactory.__init__ receives default (fallback) value as base_url.

Additionally, the list_* commands doesn't support baseurl argument

miurahr:
v3.1.10 released with the fix.

PR #732, opened 2023-11-10
Fix issue #731

comargo:
Make base_url argument optional with default to None and replace it with Settings.baseurl if it is None (standard optional argument procedure)


ddalcino (approved):
Well observed; I wrote this bug, and I'm glad you fixed it. Thank you!

PR #738, opened 2023-11-25
Improved CI to catch the problem with incorrect PRL files

iakov:
Here are problems with linkage:
https://dev.azure.com/trikset/aqtinstall/_build/results?buildId=1399&view=logs&j=c0130b29-789d-5a3c-6978-10796a508a7f&t=17fa007c-c91f-5c67-01dd-c6dc0f3947a4&l=11742

https://dev.azure.com/trikset/aqtinstall/_build/results?buildId=1399&view=logs&j=4a0dadba-b014-54f8-0bf6-a2ee97567f21&t=f7cb4e1f-77dd-5be2-be6a-0e1fc5abfb48&l=12092

This is fixed in #640  , but  (maybe) that PR should be merged into this after review.

PR #739, opened 2023-11-27
Fix *.prl files handling

iakov:
Fix problem with `*.prl` files (also patch `*.pc` files too) that contain incorrect library's path without patching. 
Seems like Qt's packages for 15.x fix this problem internally, but for some previous versions (namely, 5.11) this is a reproducible problem (as has been shown by #738)

This is a combined PR based on #738 (improved CI showing the problem) with fix from #640.
Fixes #738
Closes #640


iakov:
It really helps: https://dev.azure.com/trikset/aqtinstall/_build/results?buildId=1400&view=logs&j=4a0dadba-b014-54f8-0bf6-a2ee97567f21&t=f7cb4e1f-77dd-5be2-be6a-0e1fc5abfb48&l=12090
Compare to #738

miurahr:
Thank you!

PR #745, opened 2023-12-09
Fix azure pipeline issues

ddalcino:
There are several failures in the weekly azure pipelines CI builds, including:
* Failures while installing Qt 5.15.14 (not sure why we're installing that)
  * This PR fixes this issue
* Failures for every Qt 5.14.2-wasm_32 build that uses `emsdk@3.1.29` with `sdk-fastcomp-1.38.27-64bit`
  * I'm not certain this PR fixes this issue because I cannot trigger the pipeline run
* MacOS install of Qt 6.5.3 clang_64, without extra modules, is failing


miurahr:
#822 solved the issue.

PR #756, opened 2024-02-10
fix generate_combinations issue with Linux Qt 6.7

ddalcino:
Fix issues discovered in #751. See https://github.com/miurahr/aqtinstall/pull/751#issuecomment-1935347831

Qt 6.7.0 changes the default arch for Linux from gcc_64 to linux_gcc_64. This allows the generate_combinations.py script to account for that.

Also, Qt 6.7.0 is missing a few architectures that were present in earlier versions of Qt6. This makes sure that the script can pick those up.

### DIFF
--- a/ci/generate_combinations.py
+++ b/ci/generate_combinations.py
@@ -42,7 +42,7 @@ def iter_arches() -> Generator[dict, None, None]:
     logger.info("Fetching arches")
     archive_ids = list(iter_archive_ids(category="qt"))
     for archive_id in tqdm(archive_ids):
-        for version in ("latest", "5.15.2", "5.13.2", "5.9.9"):
+        for version in ("latest", "6.5", "6.2", "5.15.2", "5.13.2", "5.9.9"):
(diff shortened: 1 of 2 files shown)


ddalcino:
Shouldn't CI for this PR actually run the generate_combinations.py script, or did we not set it up to do that?

ci/generate_combinations.py
```diff
@@ -42,7 +42,7 @@ def iter_arches() -> Generator[dict, None, None]:
     logger.info("Fetching arches")
     archive_ids = list(iter_archive_ids(category="qt"))
     for archive_id in tqdm(archive_ids):
-        for version in ("latest", "5.15.2", "5.13.2", "5.9.9"):
+        for version in ("latest", "6.5", "6.2", "5.15.2", "5.13.2", "5.9.9"):
```
ddalcino, line 45: Oops; this doesn't actually work because 6.5 and 6.2 are not valid versions. I was counting on CI to fail if the script doesn't work; I didn't realize it wasn't going to run.

PR #765, opened 2024-03-02
Support linux_gcc_arm64 arch in 6.7.0

rectalogic:
Initial support, probably missing some cases.
Fixes #750 

Some interactive testing:
```sh-session
$ aqt list-qt linux desktop
5.9.0 5.9.1 5.9.2 5.9.3 5.9.4 5.9.5 5.9.6 5.9.7 5.9.8 5.9.9
5.10.0 5.10.1
5.11.0 5.11.1 5.11.2 5.11.3
5.12.0 5.12.1 5.12.2 5.12.3 5.12.4 5.12.5 5.12.6 5.12.7 5.12.8 5.12.9 5.12.10 5.12.11 5.12.12
5.13.0 5.13.1 5.13.2
5.14.0 5.14.1 5.14.2
5.15.0 5.15.1 5.15.2
6.0.0 6.0.1 6.0.2 6.0.3 6.0.4
6.1.0 6.1.1 6.1.2 6.1.3
6.2.0 6.2.1 6.2.2 6.2.3 6.2.4
6.3.0 6.3.1 6.3.2
6.4.0 6.4.1 6.4.2 6.4.3
6.5.0 6.5.1 6.5.2 6.5.3
6.6.0 6.6.1 6.6.2
6.7.0 6.7.0

$ aqt list-qt linux desktop --arch 6.7.0
linux_gcc_64 linux_gcc_arm64

$ aqt list-qt linux desktop --arch 6.6.0
gcc_64 wasm_singlethread wasm_multithread

$ aqt install-qt linux desktop 6.7.0 linux_gcc_arm64 --outputdir /tmp/Qt
[...]
INFO    : Finished installation of qtsvg-Linux-Debian_11_6-GCC-Linux-Debian_11_6-AARCH64.7z in 1.31891796
INFO    : Finished installation of qttools-Linux-Debian_11_6-GCC-Linux-Debian_11_6-AARCH64.7z in 1.71985637
INFO    : Downloading qttranslations...
[...]


$ aqt install-qt linux desktop 6.7.0 linux_gcc_64 --outputdir /tmp/Qt
[...]
INFO    : Finished installation of qtsvg-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 2.46250538
INFO    : Downloading qttranslations...
[...]

$ aqt install-qt linux desktop 6.6.0 gcc_64 --outputdir /tmp/Qt
[...]
[...]

Issue #748, opened 2023-12-16
Topic: deprecate offline parameter validation and "combinations.json"


pzhlkj6612:
**Is your feature request related to a problem? Please describe.**

Let's say goodbye to our old friend "Specified Qt version is unknown" and their partners. They always confuse our users.

```console
$ aqt install-qt linux desktop 9.9.9
INFO    : aqtinstall(aqt) v3.1.11 on Python 3.10.12 [CPython GCC 11.4.0]
WARNING : Specified Qt version "9.9.9" did not exist when this version of aqtinstall was released. This may not install properly, but we will try our best.
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '9.9.9'.
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt linux desktop' to show versions available.

```

```console
$ aqt install-qt linux desktop 9.9.9
INFO    : aqtinstall(aqt) v2.2.4 on Python 3.10.12 [CPython GCC 11.4.0]
WARNING : Specified Qt version is unknown: 9.9.9.
ERROR   : Failed to locate XML data for Qt version '9.9.9'.
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt linux desktop' to show versions available.

```

```console
$ aqt install 9.9.9 linux desktop
aqtinstall(aqt) v1.2.5 on Python 3.10.12 [CPython GCC 11.4.0]
Specified Qt version is unknown: 9.9.9.
Download error when access to https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/qt9_999/Updates.xml
Server response code: 404, reason: Not Found

```

**Describe the solution you'd like**

```console
$ aqt install 9.9.9 linux desktop
aqtinstall(aqt) v...
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '9.9.9'.
============================================================
Available Qt versions: 5.15.2, 6.0.4...

```

**Describe alternatives you've considered**

None.

**Additional context**

From [What's the purpose of making the "combinations.json" file? · miurahr/aqtinstall · Discussion #747](https://github.com/miurahr/aqtinstall/discussions/747).

pzhlkj6612:
Solved by #812.

pzhlkj6612:
[...]

PR #735, opened 2023-11-14
chore: update CI execution trigger/schedule

miurahr:
- Full tests weekly on master
    -  mac, windows and linux
    - Qt 5.12.12, 5.15.14, 6.5.3
    - Python 3.9, 3.10, 3.11 and 3.12
    - check sample app built  
- Change trigger for GitHub actions
    - mac, windows and linux
    - Python 3.9 and 3.12
    - Qt 5.9.9, 6.1.0
    - check qmake works

PR #766, opened 2024-03-05
Support linux_gcc_arm64 arch in 6.7.0 via new linux_arm64 host

### LINKED ISSUE #750: Issue with linux_arm64 builds
**Describe the bug**
Qt now supports linux arm64 builds via the installer
> we have added support for Linux on ARM
https://www.qt.io/blog/qt-online-installer-and-installer-framework-4.7-beta-released

https://download.qt.io/online/qtsdkrepository/linux_arm64/desktop/qt6_670/

How do I access these builds with `aqt`?

**To Reproduce**
```
$ aqt list-qt linux desktop --arch 6.7.0
linux_gcc_64
```

**Expected behavior**
Expected to see both `linux_gcc_64` and `linux_arm64`.
Actually it looks like x86 is now named `linux_x64` not `linux_gcc_64`
https://download.qt.io/online/qtsdkrepository/linux_x64/

 - OS: macOS
 - `INFO    : aqtinstall(aqt) v3.1.11 on Python 3.11.7 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]`

rectalogic:
Alternative implementation to #765 
Fixes #750

### DIFF
--- a/ci/generate_azure_pipelines_matrices.py
+++ b/ci/generate_azure_pipelines_matrices.py
@@ -92,11 +92,13 @@ python_versions = ["3.9", "3.11", "3.12"]
 qt_versions = ["5.12.12", "5.15.14", "6.5.3"]
 
 linux_build_jobs = []
+linux_arm64_build_jobs = []
 mac_build_jobs = []
 windows_build_jobs = []
 
 all_platform_build_jobs = [
     PlatformBuildJobs("linux", linux_build_jobs),
+    PlatformBuildJobs("linux_arm64", linux_arm64_build_jobs),
     PlatformBuildJobs("mac", mac_build_jobs),
     PlatformBuildJobs("windows", windows_build_jobs),
 ]
@@ -106,6 +108,7 @@ for qt_version in qt_versions:
     linux_build_jobs.append(
         BuildJob("install-qt", qt_version, "linux", "desktop", "gcc_64", "gcc_64")
     )
+linux_arm64_build_jobs.append(BuildJob("install-qt", "6.7.0", "linux_arm64", "desktop", "linux_gcc_arm64", "gcc_arm64"))
 
 # Mac Desktop
 for qt_version in qt_versions:
@@ -368,32 +371,36 @@ windows_build_jobs.extend(
 )
 
 # Test binary patch of qmake
-linux_build_jobs.extend(
-    [
-        # New output dir is shorter than the default value; qmake could fail to
-        # locate prefix dir if the value is patched wrong
-        BuildJob(
-            "install-qt",
-            "5.12.11",
-            "linux",
-            "desktop",
-            "gcc_64",
-            "gcc_64",
-            output_dir="/t/Q",
-        ),
-        # New output dir is longer than the default value.
-        # This case is meant to work without any bugfix; if this fails, the test is setup wrong
-        BuildJob(
-            "install-qt",
-            "5.12.11",
-            "linux",
-            "desktop",
-            "gcc_64",
-            "gcc_64",
-            output_dir="/some/super/long/arbitrary/path/to" * 5,
-        ),
-    ]
-)
+for lst, os_name, version, arch, arch_dir in (
+    (linux_build_jobs, "linux", "5.12.11", "gcc_64", "gcc_64"),
+    (linux_arm64_build_jobs, "linux_arm64", "6.7.0", "linux_gcc_arm64", "gcc_arm64"),
+):
+    lst.extend(
+        [
+            # New output dir is shorter than the default value; qmake could fail to
+            # locate prefix dir if the value is patched wrong
+            BuildJob(
+                "install-qt",
+                version,
+                os_name,
+                "desktop",
+                arch,
+                arch_dir,
+                output_dir="/t/Q",
+            ),
+            # New output dir is longer than the default value.
+            # This case is meant to work without any bugfix; if this fails, the test is setup wrong
+            BuildJob(
+                "install-qt",
+                version,
+                os_name,
+                "desktop",
+                arch,
+                arch_dir,
+                output_dir="/some/super/long/arbitrary/path/to" * 5,
+            ),
+        ]
+    )
 
 qt_creator_bin_path = "./Tools/QtCreator/bin/"
 qt_creator_mac_bin_path = "./Qt Creator.app/Contents/MacOS/"
@@ -479,6 +486,9 @@ for platform_build_job in all_platform_build_jobs:
 print("Setting Variables below")
 print(
     f"##vso[task.setVariable variable=linux;isOutput=true]{json.dumps(matrices['linux'])}"
+)
+print(
+    f"##vso[task.setVariable variable=linux_arm64;isOutput=true]{json.dumps(matrices['linux_arm64'])}"
 )
 print(
     f"##vso[task.setVariable variable=windows;isOutput=true]{json.dumps(matrices['windows'])}"
--- a/ci/generate_azure_pipelines_matrices.py
+++ b/ci/generate_azure_pipelines_matrices.py
@@ -371,36 +371,32 @@ windows_build_jobs.extend(
 )
 
 # Test binary patch of qmake
-for lst, os_name, version, arch, arch_dir in (
-    (linux_build_jobs, "linux", "5.12.11", "gcc_64", "gcc_64"),
-    (linux_arm64_build_jobs, "linux_arm64", "6.7.0", "linux_gcc_arm64", "gcc_arm64"),
-):
-    lst.extend(
-        [
-            # New output dir is shorter than the default value; qmake could fail to
-            # locate prefix dir if the value is patched wrong
-            BuildJob(
-                "install-qt",
-                version,
-                os_name,
-                "desktop",
-                arch,
-                arch_dir,
-                output_dir="/t/Q",
-            ),
-            # New output dir is longer than the default value.
-            # This case is meant to work without any bugfix; if this fails, the test is setup wrong
-            BuildJob(
-                "install-qt",
-                version,
-                os_name,
-                "desktop",
-                arch,
-                arch_dir,
-                output_dir="/some/super/long/arbitrary/path/to" * 5,
-            ),
-        ]
-    )
+linux_build_jobs.extend(
+    [
+        # New output dir is shorter than the default value; qmake could fail to
+        # locate prefix dir if the value is patched wrong
+        BuildJob(
+            "install-qt",
+            "5.12.11",
+            "linux",
+            "desktop",
+            "gcc_64",
+            "gcc_64",
+            output_dir="/t/Q",
+        ),
+        # New output dir is longer than the default value.
+        # This case is meant to work without any bugfix; if this fails, the test is setup wrong
+        BuildJob(
+            "install-qt",
+            "5.12.11",
+            "linux",
+            "desktop",
+            "gcc_64",
+            "gcc_64",
+            output_dir="/some/super/long/arbitrary/path/to" * 5,
+        ),
+    ]
+)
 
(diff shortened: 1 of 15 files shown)


ddalcino:
This is probably a better solution than #765. I'm sure it will be easier to maintain.

However, a few things will need to be addressed:
* Update tests: Currently, none of our tests are doing anything to test that the `linux_arm64` host works on any of the `aqt list-*` or `install-*` commands. It would be bad to merge this without adding such tests.
* The `ci/generate_combinations.py` script doesn't know anything about `linux_arm64`, and it won't be able to make automatic changes to `aqt/combinations.json` on its own. Personally, I think it makes more sense to accept your `aqt/combinations.json` as-is than update the script, and just wait for resolution of #748, which should make `combinations.json` and `generate_combinations.py` unnecessary in the first place.
* I'd really like to test this with the Azure pipelines, to make sure it didn't break anything. However, the pipelines appear to be broken right now and since #735 I'm not sure how to trigger them to run.
* I'd also like to see an automated end-to-end test, like the existing tests in the Azure pipeline, that installs Qt for arm and compiles a simple project. Whether we want to do that in GH Actions or an Azure pipeline is an open question.

rectalogic:
>     * Update tests: Currently, none of our tests are doing anything to test that the `linux_arm64` host works on any of the `aqt list-*` or `install-*` commands. It would be bad to merge this without adding such tests.

Added some tests (and some fixes for bugs the tests uncovered).

>     * I'd really like to test this with the Azure pipelines, to make sure it didn't break anything. However, the pipelines appear to be broken right now and since [chore: update CI execution trigger/schedule #735](https://github.com/miurahr/aqtinstall/pull/735) I'm not sure how to trigger them to run.

I added Azure pipeline support, using a docker image running emulated arm64. But this is all untested since I don't know how to trigger it either...

ci/generate_azure_pipelines_matrices.py
```diff
@@ -92,11 +92,13 @@ def __init__(self, platform, build_jobs):
 qt_versions = ["5.12.12", "5.15.14", "6.5.3"]
 
 linux_build_jobs = []
+linux_arm64_build_jobs = []
 mac_build_jobs = []
 windows_build_jobs = []
 
 all_platform_build_jobs = [
     PlatformBuildJobs("linux", linux_build_jobs),
+    PlatformBuildJobs("linux_arm64", linux_arm64_build_jobs),
```
miurahr, line 101: I don't think it is valuable  test it for all versions of targets.

ci/generate_azure_pipelines_matrices.py
```diff
@@ -368,32 +371,36 @@ def __init__(self, platform, build_jobs):
...
-            "desktop",
-            "gcc_64",
-            "gcc_64",
-            output_dir="/some/super/long/arbitrary/path/to" * 5,
-        ),
-    ]
-)
+for lst, os_name, version, arch, arch_dir in (
+    (linux_build_jobs, "linux", "5.12.11", "gcc_64", "gcc_64"),
+    (linux_arm64_build_jobs, "linux_arm64", "6.7.0", "linux_gcc_arm64", "gcc_arm64"),
```
miurahr, line 376: Why you test linux_arm64 with long path test case?

miurahr (changes requested):

ci/generate_azure_pipelines_matrices.py
rectalogic: I think this test is patching the `qmake` binary and so I thought there may be differences in patching an amd64 binary vs arm64. I can remove it if you think it's redundant.

ci/generate_azure_pipelines_matrices.py
rectalogic: I'm not clear how you want to limit it - should `PlatformBuildJobs` take a subset of python versions or something? It looks like `PlatformBuildJobs` is how we get variables into matrices which is used in `azure-pipelines.yml` to actually run tests via `steps.yml`? I'm not really familiar with azure pipelines so I was just following how the other platforms work, and linux_arm64 is a new platform.

ci/generate_azure_pipelines_matrices.py
ddalcino: @rectalogic please see #257, #258, and #259 for context regarding these tests. IMHO, the existing tests are already sufficient for what they’re intended to test, and I don’t see a reason to add anything more here. I appreciate that you’re trying to be thorough though.

The reality is that Azure compute time isn’t cheap, and we don’t want to waste it on tests that won’t tell us much new information. Otherwise, we would be testing this property on Mac and Windows as well.

ddalcino:
I forgot to mention, our documentation will need to be updated as well. So far, all the docs mention that the 'host' argument is only allowed to be Mac, Windows, or Linux; this won't be true anymore.

It's probably worthwhile to explain somewhere why Linux is now split between two different hosts: `linux` and `linux_arm64`, and why that isn't the case for Mac and Windows. As I understand it, after a certain version of Qt 6 (6.2 maybe?), all the Mac Qt binaries became 'universal' binaries that include both x64 and arm64. I could be wrong about that; it's been a while since I looked.

I'm not sure if there are any remaining places in the docs, or CLI help text messages, that refer to 'host' and 'operating system' interchangeably, but if there are, that should probably be cleaned up. Do `linux` and `linux_arm64` count as different operating systems? IMHO, 'host' expresses the tuple [OS, processor architecture], and that's distinct from our 'architecture' parameter, which expresses the tuple [compiler, processor architecture]. It's a bit messy, and deserves a clear explanation of what all these terms mean.

ci/generate_azure_pipelines_matrices.py
miurahr: Sorry for my misunderstanding.  My comment here is incorrect.

> I'm not clear how you want to limit it - should `PlatformBuildJobs` take a subset of python versions or something? It looks like `PlatformBuildJobs` is how we get variables into matrices which is used in `azure-pipelines.yml` to actually run tests via `steps.yml`? I'm not really familiar with azure pipelines so I was just following how the other platforms work, and linux_arm64 is a new platform.

A code chunk
```python
print("Setting Variables below")
print(
    f"##vso[task.setVariable variable=linux;isOutput=true]{json.dumps(matrices['linux'])}"
)
```
defines output matrix variable "linux"  that passed to `azure-pipelines.yml` 
then Linux tasks are defined as
```yaml
- job: Linux
  dependsOn: MatricesGenerator
  pool:
    vmImage: 'ubuntu-22.04'
  strategy:
    matrix: $[ dependencies.MatricesGenerator.outputs['mtrx.linux'] ]
  steps:
    - template: ci/steps.yml
```

Here it refers MatirxGenerator output `linux` variable.
The actual steps are defined in `ci/steps.yml`

 `PlatformBuildJobs("linux_arm64"` defines a definitions of platform "linux_arm64" and output as json.
A defined jobs will be referenced like

```yaml
  strategy:
    matrix: $[ dependencies.MatricesGenerator.outputs['mtrx.linux_arm64'] ]
```
 in `azure-pipelines.yml`.

rectalogic: > Sorry for my misunderstanding. My comment here is incorrect.

I'm not clear if you are still asking for changes or not.

> A code chunk
> 
> ```python
> print("Setting Variables below")
> print(
>     f"##vso[task.setVariable variable=linux;isOutput=true]{json.dumps(matrices['linux'])}"
> )
> ```
> 
> defines output matrix variable "linux" that passed to `azure-pipelines.yml` 

I am doing this for linux_arm64 here https://github.com/rectalogic/aqtinstall/blob/b51d200c32fd948f6e918c5ecada1dc621e68161/ci/generate_azure_pipelines_matrices.py#L486-L488

> then Linux tasks are defined as
> 
> ```yaml
> - job: Linux
>   dependsOn: MatricesGenerator
>   pool:
>     vmImage: 'ubuntu-22.04'
>   strategy:
>     matrix: $[ dependencies.MatricesGenerator.outputs['mtrx.linux'] ]
>   steps:
>     - template: ci/steps.yml
> ```

I am doing this for linux_arm64 here https://github.com/rectalogic/aqtinstall/blob/b51d200c32fd948f6e918c5ecada1dc621e68161/azure-pipelines.yml#L57-L67
This is where I use a docker container on azure to run arm64 platform (not sure this will work, needs testing)


> `PlatformBuildJobs("linux_arm64"` defines a definitions of platform "linux_arm64" and output as json. A defined jobs will be referenced like
> 
> ```yaml
>   strategy:
>     matrix: $[ dependencies.MatricesGenerator.outputs['mtrx.linux_arm64'] ]
> ```
> 
> in `azure-pipelines.yml`.

https://github.com/rectalogic/aqtinstall/blob/b51d200c32fd948f6e918c5ecada1dc621e68161/azure-pipelines.yml#L65

miurahr (approved):

ci/generate_azure_pipelines_matrices.py
miurahr: I see. Now I can understand and LGTM.

rectalogic:
6.7 is released
https://www.qt.io/blog/qt-6.7-released

Should we merge this and do a new aqtinstall release?

rectalogic:
@miurahr can you merge this? Currently even amd64 linux 6.7.0 is not installable with aqt 3.1.12 even if you use the new arch name (`linux_gcc_64` instead of `gcc_64`):

```sh-session
> aqt install-qt linux desktop 6.7.0 -O /tmp/qt
INFO    : aqtinstall(aqt) v3.1.12 on Python 3.11.8 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]
ERROR   : The packages ['qt_base'] were not found while parsing XML of package information!
* Please use 'aqt list-qt linux desktop --arch 6.7.0' to show architectures available.

> aqt install-qt linux desktop 6.7.0 linux_gcc_64 -O /tmp/qt
INFO    : aqtinstall(aqt) v3.1.12 on Python 3.11.8 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]
INFO    : Downloading qtsvg...
INFO    : Downloading qttools...
INFO    : Downloading qtdeclarative...
INFO    : Downloading qtbase...
INFO    : Redirected: qt.mirror.constant.com
INFO    : Redirected: qt.mirror.constant.com
INFO    : Redirected: qt.mirror.constant.com
INFO    : Redirected: qt.mirror.constant.com
INFO    : Finished installation of qtsvg-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 1.18460154
INFO    : Downloading qttranslations...
INFO    : Redirected: qt.mirror.constant.com
INFO    : Finished installation of qttranslations-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 1.32922746
INFO    : Downloading qtwayland...
INFO    : Redirected: qt.mirror.constant.com
INFO    : Finished installation of qtwayland-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 1.12197392
INFO    : Downloading icu...
INFO    : Finished installation of qtbase-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 4.41483796
INFO    : Redirected: qt.mirror.constant.com
INFO    : Finished installation of qttools-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 5.18647496
INFO    : Finished installation of icu-linux-Rhel8.6-x86_64.7z in 2.62416004
INFO    : Finished installation of qtdeclarative-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 11.03515671
ERROR   : Updater caused an IO error: [Errno 2] No such file or directory: '/tmp/qt/6.7.0/linux_gcc_64/mkspecs/qconfig.pri'
```

Issue #769, opened 2024-03-17
Fails to install Qt6 6.7.0 on Linux/Ubuntu


MehdiChinoune:
**Describe the bug**
Installing Qt6 6.7.0 (RC) succeeds on Windows and macOS but fails on Ubuntu

**To Reproduce**
1. install `aqt` using `pip install aqtinstall`
2. run `aqt` with this command `aqt install-qt linux desktop 6.7.0`
3. See error

**Expected behavior**
It succeeds

**`aqt` output**

```
INFO    : aqtinstall(aqt) v3.1.12 on Python 3.10.13 [CPython GCC 11.4.0]
ERROR   : The packages ['qt_base'] were not found while parsing XML of package information!
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt linux desktop --arch 6.7.0' to show architectures available.
```

**Desktop (please complete the following information):**
 - OS: Linux/Ubuntu 22.04
 - `aqt` version 3.1.12

MehdiChinoune:
Executing with arch succeeds
```
aqt install-qt linux desktop 6.7.0 linux_gcc_64
```

miurahr:
> because that file is in gcc_64 not in linux_gcc_64

Uhm, ` find_installed_desktop_qt_dir` function that returns the directory checks existence of `bin/qmake` command.
https://github.com/miurahr/aqtinstall/blob/master/aqt/metadata.py#L470

but `update` function
https://github.com/miurahr/aqtinstall/blob/master/aqt/updater.py#L269
does not take a prefix from the `find_installed_desktop_qt_dir` function.
https://github.com/miurahr/aqtinstall/blob/master/aqt/updater.py#L288-L289

This may be one of causes.

miurahr:
> > because that file is in gcc_64 not in linux_gcc_64
>
>  https://github.com/miurahr/aqtinstall/blob/master/aqt/updater.py#L288-L289

`arch_dir` is found by using `QtRepoProperty.get_arch_dir_name`

```
        arch_dir = QtRepoProperty.get_arch_dir_name(os_name, arch, version)
```
that just return `arch`  variable string when host is `linux`.
I can check `arch` is `linux_gcc_64` and then return `gcc_64`.
We can also look an existence of `qmake` file here for double check.

firewave:
Qt 6.7.0 was released today and I just hit this issue via the `jurplel/install-qt-action` GitHub action.

timangus:
[...]

PR #770, opened 2024-03-20
fix: always use gcc_64 as arch_dir for linux

miurahr:
From Qt 6.7.0, arch keyword is changed to linux_gcc_64 but arch_dir should always be gcc_64. 
The change fixes issue #769 Fails to install Qt6 6.7.0 on Linux/Ubuntu


rectalogic:
FYI this is also fixed in https://github.com/miurahr/aqtinstall/pull/766/files#diff-b336135f575fd47d95d02ec92ce8e4fa51a5add90b3d2057af8ecad46b204891R388

That fix also accounts for the new `gcc_arm64` arch dir.

Issue #775, opened 2024-04-05
install-src and install-doc fail for 6.7.0


rectalogic:
**Describe the bug**
install-src and install-doc both fail for macos

**To Reproduce**
```sh-session
> aqt install-src mac 6.7.0 -O /tmp/qt
INFO    : aqtinstall(aqt) v3.1.12 on Python 3.11.8 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '6.7.0'.
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt mac desktop' to show versions available.
> aqt install-doc mac 6.7.0 -O /tmp/qt
INFO    : aqtinstall(aqt) v3.1.12 on Python 3.11.8 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '6.7.0'.
==============================Suggested follow-up:==============================
* Please use 'aqt list-qt mac desktop' to show versions available.
```

**Expected behavior**
Should install

**`aqt` output**
See attached [aqtinstall.log](https://github.com/miurahr/aqtinstall/files/14888604/aqtinstall.log)

**Desktop (please complete the following information):**
 - OS: macOS 14.4.1
 - `INFO    : aqtinstall(aqt) v3.1.12 on Python 3.11.8 [CPython Clang 15.0.0 (clang-1500.1.0.2.5)]`

**Additional context**
`aqt` is attempting to download https://download.qt.io/online/qtsdkrepository/mac_x64/desktop/qt6_670_src_doc_examples/Updates.xml.sha256 but it appears 6.7.0 has moved this to https://download.qt.io/online/qtsdkrepository/all_os/qt/qt6_670_src_doc_examples/Updates.xml

pcolby:
The same issue applies to Linux too (but not Windows).

```
2024-04-29 00:04:10,247 - aqt.main - INFO - installer 139878233721728 aqtinstall(aqt) v3.1.13 on Python 3.12.3 [CPython GCC 11.4.0]
[...]

PR #776, opened 2024-04-05
Override host/target for src/docs if Qt >= 6.7.0

rectalogic:
Fixes https://github.com/miurahr/aqtinstall/issues/775

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -482,7 +482,7 @@ class Cli:
             qt_version = args.qt_version
             Cli._validate_version_str(qt_version)
         # Override target/os for recent Qt
-        if Version(qt_version) >= Version("6.7.0"):
+        if Version(qt_version) in SimpleSpec(">=6.7.0"):
             target = "qt"
             os_name = "all_os"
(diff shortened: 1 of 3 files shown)


aqt/installer.py
```diff
@@ -481,6 +481,10 @@ def _run_src_doc_examples(self, flavor, args, cmd_name: Optional[str] = None):
         else:
             qt_version = args.qt_version
             Cli._validate_version_str(qt_version)
+        # Override target/os for recent Qt
+        if Version(qt_version) >= Version("6.7.0"):
```
miurahr, line 485: This can be `if Version(qt_version) in SimpleSpec(">=6.7.0"):`

miurahr: or `if match(">=6.7.0", qt_version):` 

https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.match

We use `version-in-spec` syntax in metadata.py, so it seems better to use similar one.

miurahr:
@rectalogic Thanks!

PR #801, opened 2024-06-21
fix: installation of android for Qt 6.8.0

miurahr:
It modified a URL path of  android distribution for Qt 6.8.0 and later.

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -362,7 +362,11 @@ class QtArchives:
         return target_packages
 
     def _get_archives(self):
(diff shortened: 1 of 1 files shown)


aqt/archives.py
```diff
@@ -362,7 +362,11 @@ def _target_packages(self) -> ModuleToPackage:
         return target_packages
 
     def _get_archives(self):
-        self._get_archives_base(f"qt{self.version.major}_{self._version_str()}{self._arch_ext()}", self._target_packages())
+        if self.version >= Version("6.8.0"):
+            name = f"qt{self.version.major}_{self._version_str()}/qt{self.version.major}_{self._version_str()}{self._arch_ext()}"
```
miurahr, line 366: Line is too long here.

Issue #807, opened 2024-07-28
[Bug]:  CI failed: combination detection 


miurahr:
### Bug description

CI `combination.json` failed 

### Expected behavior

Run CI without error

### aqt and python version

Python 3.12

### Operating System

Linux/Unix

### Relevant log output

```shell
INFO    : Fetching arches
Traceback (most recent call last):
  File "/home/runner/work/aqtinstall/aqtinstall/ci/generate_combinations.py", line 215, in <module>
    main(filename=json_filename, is_write_file=args.write, is_verbose=args.verbose)
  File "/home/runner/work/aqtinstall/aqtinstall/ci/generate_combinations.py", line 147, in main
    actual = [generate_combos(new_archive=expect[0]["new_archive"])]
  File "/home/runner/work/aqtinstall/aqtinstall/ci/generate_combinations.py", line 120, in generate_combos
    "qt": merge_records(iter_arches()),
  File "/home/runner/work/aqtinstall/aqtinstall/ci/generate_combinations.py", line 108, in merge_records
    for record in arch_records:
  File "/home/runner/work/aqtinstall/aqtinstall/ci/generate_combinations.py", line 48, in iter_arches
    for arch_name in MetadataFactory(archive_id, architectures_ver=version).getList():
  File "/home/runner/work/aqtinstall/aqtinstall/aqt/metadata.py", line 620, in getList
    return self._action()
  File "/home/runner/work/aqtinstall/aqtinstall/aqt/metadata.py", line 600, in <lambda>
    self._action = lambda: self.fetch_arches(self._to_version(ver_str, None))
  File "/home/runner/work/aqtinstall/aqtinstall/aqt/metadata.py", line 720, in _to_version
    raise CliInputError(msg)
aqt.exceptions.CliInputError: There is no latest version of Qt with the criteria 'qt/all_os/qt'
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

pzhlkj6612:
Is this the chance to drop the offline validation? #748.

miurahr:
Lets do it!

pzhlkj6612:
Is there something I can help?

miurahr:
I think we can remove lines https://github.com/miurahr/aqtinstall/blob/master/aqt/installer.py#L420-L423 and related methods.  Could you try the modification? If it works, it is a chance to raise Pull-Request.

miurahr:
#812

pzhlkj6612:
[...]

PR #812, opened 2024-08-04
fix: delete combination check

miurahr:
Checking Qt version and modules combination become unfeasible, because release of new versions are frequent.
Also aqtinstall now have a feature to list versions, modules, archives.  

This is a proposal of removal of combination check feature.

- Delete combinations.json and relevant CI script to update it
- Delete feature to print warning with check combinations
- Delete Settings class methods and field to examine combination database
- Remove tests to check combinations
- Update expectation of tests, which expect showing warn.


pzhlkj6612:
Will close:

- #748 
- #807

.

docs/CHANGELOG.rst
```diff
@@ -13,6 +13,10 @@ All notable changes to this project will be documented in this file.
 `Unreleased`_
 =============
 
+Deprecated
+----------
+* Qt version/architecture/modules checking (#807, #812)
+
```
pzhlkj6612, line 19: Deprecated? I think a deprecated function is still available. We just deleted it, so I suggest using a new word. Which word?

In addition:

```suggestion
Deprecated
----------
* Qt version/architecture/modules checking (#748, #807, #812)

```

.

pzhlkj6612 (changes requested):

docs/CHANGELOG.rst
miurahr: I would like to use "Deleted"

docs/CHANGELOG.rst
```diff
@@ -13,6 +13,10 @@ All notable changes to this project will be documented in this file.
 `Unreleased`_
 =============
 
+Deprecated
+----------
```
miurahr, line 17:
```suggestion
Deleted
-------
```

pzhlkj6612 (approved):
lgtm

PR #826, opened 2024-10-05
bug: fix error when instaling win64_llvm_mingw / Qt 6.7 and 6.8

miurahr:
ref #825

Issue #825, opened 2024-10-05
[Bug]: Install fails for win64_llvm_mingw arch


pcolby:
### Bug description

For Qt 6.7 (and 6.8) it should be possible to install the `win64_llvm_mingw` arch, however, it appears that `aqt` expects to find the `qconfig.pri` in an incorrect location eg:

```
ERROR   : Updater caused an IO error: [Errno 2] No such file or directory: '/home/paul/tmp/aqt/6.7.3/llvm_mingw/mkspecs/qconfig.pri'
```

But the `qconfig.pri` file is under a `llvm-mingw_64` directory, not `llvm_mingw`.

### Expected behavior

This should work, without error:
```
aqt install-qt windows desktop 6.7.3 win64_llvm_mingw
aqt install-qt windows desktop 6.8.0 win64_llvm_mingw
```

Thanks!

### aqt and python version

v3.1.17 on Python 3.10.4 [CPython GCC 13.2.0]

### Operating System

Linux/Unix, Windows

### Relevant log output

```shell
paul@jafar:~/tmp/aqt$ aqt install-qt windows desktop 6.7.3 win64_llvm_mingw
INFO    : aqtinstall(aqt) v3.1.17 on Python 3.10.4 [CPython GCC 13.2.0]
WARNING : Specified Qt version "6.7.3" did not exist when this version of aqtinstall was released. This may not install properly, but we will try our best.
INFO    : Downloading qtdeclarative...
INFO    : Downloading qtsvg...
INFO    : Downloading qtbase...
INFO    : Downloading qttools...
INFO    : Redirected: mirror.aarnet.edu.au
INFO    : Redirected: mirror.aarnet.edu.au
INFO    : Redirected: mirror.aarnet.edu.au
INFO    : Redirected: mirror.aarnet.edu.au
INFO    : Finished installation of qtsvg-Windows-Windows_11_23H2-Clang-Windows-Windows_11_23H2-X86_64.7z in 3.12440094
INFO    : Finished installation of qttools-Windows-Windows_11_23H2-Clang-Windows-Windows_11_23H2-X86_64.7z in 4.44455947
INFO    : Downloading qttranslations...
INFO    : Downloading llvm...
INFO    : Redirected: mirror.aarnet.edu.au
INFO    : Finished installation of qttranslations-Windows-Windows_11_23H2-Clang-Windows-Windows_11_23H2-X86_64.7z in 2.88042200
[...]

PR #829, opened 2024-10-06
Attempted fix for CI error in #826

ddalcino:


jdpurcell:
I reported a [similar issue](https://github.com/miurahr/aqtinstall/issues/825#issuecomment-2400125218) with win64_msvc2022_arm64_cross_compiled. Not sure if it's closely enough related to warrant being in the same PR, but I'd imagine the fix is somewhat similar, i.e. adding another special case in `get_arch_dir_name`.
```patch
         elif arch.startswith("win"):
             m = re.match(r"win\d{2}_(?P<msvc>msvc\d{4})_(?P<winrt>winrt_x\d{2})", arch)
             if m:
                 return f"{m.group('winrt')}_{m.group('msvc')}"
+            elif arch.endswith("_cross_compiled"):
+                return arch[6:-15]
             else:
                 return arch[6:]
```
Like these two new lines maybe?

ddalcino:
> I reported a [similar issue](https://github.com/miurahr/aqtinstall/issues/825#issuecomment-2400125218) with win64_msvc2022_arm64_cross_compiled. Not sure if it's closely enough related to warrant being in the same PR, but I'd imagine the fix is somewhat similar, i.e. adding another special case in `get_arch_dir_name`.

 You are correct; this is something that does need to change. Until today, I was not aware that `win64_msvc2022_arm64_cross_compiled` was a valid architecture, but here it is:

```shell
$ python -m aqt list-qt windows desktop --arch 6.8.0
win64_llvm_mingw win64_mingw win64_msvc2022_64 win64_msvc2022_arm64_cross_compiled
```

In any case, I think this is outside the scope of my PR. This PR is just for fixing an error in CI; I don't want to get into supporting new architectures here.

Issue #820, opened 2024-09-25
[Question]: SSL issue on machine with self signed root certificate


dk-mushiyoke:
### Bug description

Running `python -m aqt install-qt --outputdir ./deps/qt windows desktop 6.7.2 win64_msvc2019_64` on a machine that has a self signed root cert that routes all traffic to a proxy results in SSL error on all mirrors.

The mirrors work fine in a browser as the root cert has been trusted in Windows Certificate Manager

### Expected behavior

Would be nice to be able to pass a certificate to the command, or have a flag that ignores SSL errors

### aqt and python version

aqtinstall(aqt) v3.1.18 on Python 3.12.6 [CPython MSC v.1940 64 bit (AMD64)]

### Operating System

Windows

### Relevant log output

```shell
INFO    : aqtinstall(aqt) v3.1.18 on Python 3.12.6 [CPython MSC v.1940 64 bit (AMD64)]
WARNING : Specified Qt version "6.7.2" did not exist when this version of aqtinstall was released. This may not install properly, but we will try our best.
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
WARNING : Connection to 'https://download.qt.io' failed. Retrying with fallback 'https://qt.mirror.constant.com/'.
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failure to connect to https://qt.mirror.constant.com/online/qtsdkrepository/windows_x86/desktop/qt6_672/Updates.xml: SSLError
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

miurahr:
Requests, a HTTP network access library that aqtinstall depends on, does not use the Windows Certification Manager.
Please consult with the requests manual.

- [Requests - manual](https://requests.readthedocs.io/en/latest/)
- [Requests - manual - CA Certificate](https://requests.readthedocs.io/en/latest/user/advanced/#ca-certificates)

dk-mushiyoke:
@miurahr thank you for the links to Requests' documentation site, I was able to follow [this section](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification) and `set REQUESTS_CA_BUNDLE=path/to/pem` environment variable. aqt then was able to download and install normally.

PR #830, opened 2024-10-07
docs: add appendix to explain special case.

miurahr:
Inspired by question #820 
it is valuable to help people in special environment


docs/appendix.md
```diff
@@ -0,0 +1,50 @@
+.. _appendix-ref:
+
+Appendix
+========
+
+Hint to avoid netowrk errors
```
ddalcino, line 6:
```suggestion
Hint to avoid network errors
```

miurahr: Thanks. I've fixed the typo.

PR #841, opened 2024-11-30
handle Updates.xml Operation extract elements.

### LINKED ISSUE #840: [Bug]: 6.8.1 installations fail, improper archive unpacking
### Bug description

In 6.8.1 Updates.xml files PackageUpdate elements can contain an Operations element.  These can contain Operation elements with attribute Extract, which in turn contain Argument elements that contain the target directory for extraction as well as the name of the 7z file.

For example see https://[download.qt.io/online/qtsdkrepository/linux_x64/desktop/qt6_681/qt6_681/Updates.xml](https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/qt6_681/qt6_681/Updates.xml)

aqtinstall doesn't handle these optional elements, instead unpacking everything assuming the path names are stored in the 7z file.  This is no longer true for 7z archives in 6.8.1.

`aqt install-qt linux desktop 6.8.1 linux_gcc_64 -O /tmp/Qt -m qt3d qt5compat qtconnectivity qtimageformats qtlanguageserver qtlocation qtmultimedia qtpdf qtpositioning qtquickeffectmaker qtremoteobjects qtscxml qtsensors qtserialbus qtserialport qtspeech qtwebchannel qtwebengine qtwebsockets qtwebview`
leads to
```ERROR   : Updater caused an IO error: [Errno 2] No such file or directory: '/tmp/Qt/6.8.1/gcc_64/mkspecs/qconfig.pri'```
This would seem to be caused by unpacking all archives in the wrong place.   All 7z archives are unpacked to the directory supplied in the command line, missing at a minimum the usual version/compiler/ path component, in this case we would have expected everything to go to /tmp/Qt/6.8.1/gcc_64/.  Thee missing pathname components appear to all be given in <Operation name="Extract"> Argument elements.  Note these missing pathname components may contain additional subdirectories besides the version/compiler components, for example icu-linux-Rhel8.6-x86_64.7z should have been unpacked to @TargetDir@/6.8.1/gcc_64/lib.
```
$ ls /tmp/Qt
bin      lib               libicudata.so.73.2  libicuio.so.73    libicutest.so.73.2  libicuuc.so.73    mkspecs      plugins    sbom
doc      libexec           libicui18n.so.73    libicuio.so.73.2  libicutu.so.73      libicuuc.so.73.2  modules      qml        translations
include  libicudata.so.73  libicui18n.so.73.2  libicutest.so.73  libicutu.so.73.2    metatypes         phrasebooks  resources
```

Note I was testing with code from #838 but this should be reproducible with released code if one omits the modules qtwebengine and qtpdf which are extensions in 6.8 (#803).

### Expected behavior

A correct directory structure, in the example in /tmp/Qt/6.8.1/gcc_64.

### aqt and python version

 aqtinstall(aqt) v1.2.0b3.dev1123 on Python 3.12.3 [CPython GCC 13.2.0]

### Operating System

Linux/Unix

### Relevant log output

```shell
2024-11-30 06:56:17,513 - aqt.main - ERROR - installer 139666317992064 Updater caused an IO error: [Errno 2] No such file or directory: '/tmp/Qt/6.8.1/gcc_64/mkspecs/qconfig.pri'
2024-11-30 07:28:08,466 - aqt.main - INFO - installer 140246693441664 aqtinstall(aqt) v1.2.0b3.dev1123 on Python 3.12.3 [CPython GCC 13.2.0]
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tsteven4:
This resolves #840

### DIFF
--- a/.github/workflows/test-install-qt.yml
+++ b/.github/workflows/test-install-qt.yml
@@ -31,6 +31,11 @@ jobs:
             py: "3.12"
             qtver: 6.8.0
             artifact: standard
+          - os: ubuntu-latest
+            py: "3.12"
+            # 6.8.1 introduces Operations elements in Updates.xml
+            qtver: 6.8.1
+            artifact: standard
     steps:
       - uses: actions/checkout@v4
         with:
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -190,7 +190,7 @@ class Updates:
             dependencies = updates._get_list(packageupdate.find("Dependencies"))
             auto_dependon = updates._get_list(packageupdate.find("AutoDependOn"))
             archives = updates._get_list(packageupdate.find("DownloadableArchives"))
-            archive_install_paths = []
+            archive_install_paths = updates._get_list(None)
             default = updates._get_boolean(packageupdate.find("Default"))
             virtual = updates._get_boolean(packageupdate.find("Virtual"))
             if packageupdate.find(extract_xpath) is not None:
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -21,6 +21,7 @@
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 import posixpath
 from dataclasses import dataclass, field
+from itertools import islice, zip_longest
 from logging import getLogger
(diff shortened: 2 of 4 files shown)


miurahr:
LGTM.

PR #846, opened 2024-12-05
Fix WASM

### LINKED ISSUE #774: Unable to install Android 6.7.0
**Describe the bug**
`aqt list-qt windows android` doesn't show 6.7.0 and it's not possible to install it. Same with linux or macos

Looks like starting with 6.7.0, the Android archives are moved to `all_os` folder, here: https://download.qt.io/online/qtsdkrepository/all_os/android/qt6_670_arm64_v8a/

**To Reproduce**
1. Run `aqt install-qt windows android 6.7.0 android_arm64_v8a`

**Expected behavior**
It should download and install the 6.7.0 Android package

 - OS: Windows 11
 - `aqtinstall(aqt) v3.1.12 on Python 3.10.11 [CPython MSC v.1929 64 bit (AMD64)]`

### LINKED ISSUE #779: Installing WebAssembly Qt 6.7.0 fails
**Describe the bug**
Qt for WebAssembly 6.7.0 fails to install as before, and doesn't appear to be available.

**To Reproduce**
1. Run `aqt install-qt linux desktop 6.7.0 wasm_multithread --autodesktop`
2. Observe error message.

**Expected behavior**
Qt for WebAssembly 6.7.0 is installed.

**`aqt` output**
```
$ aqt install-qt linux desktop 6.7.0 wasm_multithread --autodesktop
INFO    : aqtinstall(aqt) v3.1.13 on Python 3.10.12 [CPython GCC 11.4.0]
INFO    : You are installing the Qt6-WASM version of Qt, which requires that the desktop version of Qt is also installed. Now installing Qt: desktop 6.7.0 linux_gcc_64
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '6.7.0'.
* Please use 'aqt list-qt linux desktop' to show versions available.

$ aqt list-qt linux desktop --arch 6.7.0
linux_gcc_64
```

 - OS: linux
 - `aqt` version: aqtinstall(aqt) v3.1.13 on Python 3.10.12 [CPython GCC 11.4.0]

**Additional context**
Valid output of 6.6.3 query, for comparison:
```
$ aqt list-qt linux desktop --arch 6.6.3
gcc_64 wasm_singlethread wasm_multithread
```

Kidev:
### Description

- Fix `install-qt` Qt WASM packages for Qt versions 6.7.0 and above
```console
kidev:~$ aqt install-qt all_os wasm 6.8.1 wasm_singlethread --autodesktop --modules qtquick3d qtshadertools
```
- Fix `list-qt` for Qt WASM packages for Qt versions 6.7.0 and above
```console
kidev:~$ aqt list-qt all_os wasm --arch 6.7.3
wasm_singlethread wasm_multithread
```
- Handle the specificity of Qt versions 6.7.0 and above while remaining backward compatible
```console
kidev:~$ aqt install-qt windows desktop 6.7.3 win64_llvm_mingw
kidev:~$ aqt install-qt windows desktop 6.8.0 win64_llvm_mingw
kidev:~$ aqt install-qt windows android 6.7.0 android_arm64_v8a
```
- Fix crash and issues with the new Qt version notation
```python
>>> get_semantic_version('51212', False)
Version('5.12.12')
>>> get_semantic_version('600', False)
Version('6.0.0')
>>> get_semantic_version('6_7_3', False)
Version('6.7.3')
```
- Adds tests for `get_semantic_version`, WASM Qt 6.7.3+ and Qt 6.8.0+
- Make CI use Qt 6.8.0+ and test it in priority in addition to previous versions
- Make CI use a version of EMSDK & MSVC matching the Qt version being built
- Make CI use C++20 for relevant Qt versions
- Update `help` commands to match the latest version
- Added comments on obscure functions to make the code easier to follow

> [!NOTE]
> For those using [jurplel/install-qt-action](https://github.com/marketplace/actions/install-qt), until [the current PR](https://github.com/miurahr/aqtinstall/pull/846) and [this PR](https://github.com/jurplel/install-qt-action/pull/272) are merged, you can use my Action [on the market here](https://github.com/marketplace/actions/install-qt-edited), and you can use my aqtinstall fork by using the following configuration directly as is:
> ```yml
> - name: Install Qt for host architecture
>   uses: Kidev/install-qt-action@v4.2.0
>   with:
>     version: '6.8.1'
>     host: 'all_os'
>     target: 'wasm'
>     arch: 'wasm_singlethread'
>     set-env: 'true'
>     modules: 'qtquick3d'
>     aqtsource: 'git+https://github.com/Kidev/aqtinstall.git@wasm'
>     extra: '--autodesktop'
> ```
> You can check the [successful deploy with Qt WASM 6.8.1 on a project of mine here](https://github.com/Kidev/QtQuick3D-Tools/actions/runs/12447017607) that is using this config
>
> Once the PRs are merged, you'll be able to use the official one:
> ```yml
> - name: Install Qt for host and WASM
>   uses: jurplel/install-qt-action@v4.?.?
>   with:
>     version: '6.8.1'
>     host: 'all_os'
>     target: 'wasm'
>     arch: 'wasm_singlethread'
>     modules: 'all'
> ```

Fix #779 fix #774 and more


Kidev:
`get_semantic_version` is now consistent with preview versions naming:
```python
# Before PR
5.12.12
6.0.0
None
5.1212-preview
6.0-preview
None

# Before commit c4ce9f0
5.12.12
6.0.0
6.7.3
None
None
None

# Now
5.12.12
6.0.0
6.7.3
5.1212-preview
6.0-preview
6.73-preview
```

Kidev:
Check failed because of mirror timeout: it passes https://github.com/miurahr/aqtinstall/runs/34295379939

miurahr:
Could you add a test case for wasm 6.7.3 in the test code?
https://github.com/miurahr/aqtinstall/blob/master/tests/test_list.py#L426

Also, could you update the CI variation?
https://github.com/miurahr/aqtinstall/blob/master/ci/generate_azure_pipelines_matrices.py#L217
Currently checks Qt 6.4.x and 6.5.x with emsdk 3.1.x

miurahr:
Is it possible to add a unit test for the `get_semantic_version` to check values you notified?

> `get_semantic_version` is now consistent with preview versions naming, and have the correct error raised with the correct error message
> 
> ```python
> get_semantic_version('51212', False) -> '5.12.12'
> get_semantic_version('600', False) -> '6.0.0'
> get_semantic_version('6_7_3', False) -> '6.7.3'
> get_semantic_version('51212', True) -> '5.1212-preview'
> get_semantic_version('600', True) -> '6.0-preview'
> get_semantic_version('6_7_3', True) -> '6.73-preview'
> ```

The chunk you added is not tested, on test coverage visualization on coveralls.io service.
https://coveralls.io/builds/71332638/source?filename=aqt%2Fmetadata.py#L187

miurahr:
Please confirm the coverage of your additions also for `archives.py`. That is why I asked to add test case for both 6.7.3 and 6.8.0.
https://coveralls.io/builds/71332638/source?filename=aqt%2Farchives.py#L390

miurahr:
> Also, could you update the CI variation? https://github.com/miurahr/aqtinstall/blob/master/ci/generate_azure_pipelines_matrices.py#L217 Currently checks Qt 6.4.x and 6.5.x with emsdk 3.1.x

From my short research, Qt LTS under the support duration is version 6.5.x and 6.8.x. It may be better to check these two on CI.

Kidev:
Almost done. I've added tests, and most importantly completely rewrote my approach. I chose to stick to Qt's team logic rather than trying to create an impression of continuity. Since 6.7.x, Qt decided that WASM was `host=all_os` `target=wasm`, so be it:
```
$ aqt list-qt all_os wasm --arch 6.8.1 
wasm_singlethread wasm_multithread

$ aqt list-qt all_os wasm --arch 6.7.3
wasm_singlethread wasm_multithread

$ aqt install-qt all_os wasm 6.7.3 wasm_singlethread
```

It is much more consistent with the current code and logic of aqt, adds much less magic branches with random versions, and uses more of the existing code architecture.

Kidev:
Moved to latest version as described in previous comment. I'm not very familiar with Azure CI so we'll see. Otherwise, `tox` returns a 100% tests passing, and I added the tests requested.

tests/test_list.py
```diff
@@ -435,7 +478,6 @@ def expected_windows_desktop_plus_wasm_5140(is_wasm_threaded: bool) -> Dict:
         ("--modules 5.14.0 win64_msvc2017_64", False, ["modules_by_arch", "win64_msvc2017_64"]),
         ("--modules 6.5.0 wasm_singlethread", True, ["modules_by_arch", "wasm_singlethread"]),
         ("--modules 6.5.0 wasm_multithread", True, ["modules_by_arch", "wasm_multithread"]),
-        ("--arch latest", True, ["architectures"]),
```
Kidev, line 438: This used to work, but now it refers to 6.8.1, where there are less archs available when going to all_os wasm: only `wasm_singlethread` `wasm_multithread`, and this was still excepting the other standard archs. Hence the removal. Tell me if I missed something and that it should be kept

Kidev:
Most errors are due to what we discussed [here](https://github.com/miurahr/aqtinstall/pull/852#issuecomment-2544118204), I forgot that it would impact mac as well, should be an easy fix. A merge of #852 into master would also solve this one I sync.

Kidev:
Ok so it was not only the issue of the extensions. Looks like --autodesktop forces the target to be desktop and that's an issue for wasm targets, I'll fix that

aqt/archives.py
```diff
@@ -454,7 +461,7 @@ def _download_update_xml(self, update_xml_path, silent=False):
         """Hook for unit test."""
         if not Settings.ignore_hash:
             try:
-                xml_hash = get_hash(update_xml_path, Settings.hash_algorithm, self.timeout)
+                xml_hash = get_hash(update_xml_path, Settings.hash_algorithm, self.timeout, one_rep=silent)
```
Kidev, line 464: See the comment on aqt/helper.py

aqt/helper.py
```diff
@@ -210,6 +210,8 @@ def get_hash(archive_path: str, algorithm: str, timeout: Tuple[float, float]) ->
             if len(_hash) == hash_lengths[algorithm]:
                 return binascii.unhexlify(_hash)
         except (ArchiveConnectionError, ArchiveDownloadError, binascii.Incomplete, binascii.Error):
+            if one_rep:
```
Kidev, line 213: This is so that, if the extensions are checked silently, they cannot try all the mirrors. This is not the idea solution, but invert the caller would be longer.

Why is this issue? The caller is basically spamming the servers: at list-qt, it pulls the webpage for each extension. But downstream, for each of those, get_hash can pull it as many times as there is mirrors. If the download error happens because the client entered a bad arch, then this causes aqt to freeze. It tries to pull nb_extensions * nb_mirrors, with each time a timeout. Do this a few time and you get timeed out by major mirror for too many calls in a short time

aqt/helper.py
```diff
@@ -210,6 +210,8 @@ def get_hash(archive_path: str, algorithm: str, timeout: Tuple[float, float]) ->
...
+                break
```
Kidev, line 214: This is so that, if the extensions are checked silently, they cannot try all the mirrors. This is not the idea solution, but invert the caller would be longer.

Why is this issue? The caller is basically spamming the servers: at list-qt, it pulls the webpage for each extension. But downstream, for each of those, get_hash can pull it as many times as there is mirrors. If the download error happens because the client entered a bad arch, then this causes aqt to freeze. It tries to pull nb_extensions * nb_mirrors, with each time a timeout. Do this a few time and you get timeed out by major mirror for too many calls in a short time

Kidev:
One error of the CI is that Qt now requires C++17:
```console
Qt\6.8.1\msvc2022_64\include\QtCore/qcompilerdetection.h(1258): fatal error C1189: #error:  "Qt requires a C++17 compiler, and a suitable value for __cplusplus. On MSVC, you must pass the /Zc:__cplusplus option to the compiler."```

I'm going to try to fix that but as I said I'm not familiar with Azure

**Comment by miurahr** [comment]
> One error of the CI is that Qt now requires C++17:
> 
> ```
> Qt\6.8.1\msvc2022_64\include\QtCore/qcompilerdetection.h(1258): fatal error C1189: #error:  "Qt requires a C++17 compiler, and a suitable value for __cplusplus. On MSVC, you must pass the /Zc:__cplusplus option to the compiler."
> ```
> 
> I'm going to try to fix that but as I said I'm not familiar with Azure

Could you look  `ci/steps.yml` that defines steps to execute? There are MSVC related code in line 275.

```
  - powershell: |
      if ( $env:TOOLCHAIN -eq 'MSVC' ) {
        # Load modules from cache  
        $Env:PSModulePath = '$(MODULES_FOLDER)', $Env:PSModulePath -join [System.IO.Path]::PathSeparator
        Write-Host $Env:PSModulePath
        Import-Module "Pscx"
        Import-Module "VSSetup"

        Import-VisualStudioVars -VisualStudioVersion $(VSVER) -Architecture $(ARCHITECTURE)
        $env:Path += ";$(WIN_QT_BINDIR)"
        mkdir $(Build.BinariesDirectory)\tests
        cd $(Build.BinariesDirectory)\tests
        7z x $(Build.SourcesDirectory)\ci\helloworld.7z
        cd ..
        qmake $(Build.BinariesDirectory)\tests\helloworld
        jom
      } elseif ( $env:TOOLCHAIN -eq 'MINGW' ) {
        python -m aqt install-tool $(if (($QT_BASE_MIRROR + "") -ne "") { "-b $QT_BASE_MIRROR" } else {""}) `
          --outputdir $(Build.BinariesDirectory)/Qt $(HOST) desktop $(MINGW_TOOL_NAME) qt.tools.$(MINGW_VARIANT)
        if ($?) {
          Write-Host 'Successfully installed tools_mingw'
        } else {
          throw 'Failed to install tools_mingw'
        }
        Set-Item -Path Env:Path -Value ("$(Build.BinariesDirectory)\Qt\Tools\$(MINGW_FOLDER)\bin;$(WIN_QT_BINDIR);" + $Env:Path)
        Write-Host "Path == " + $env:Path
        mkdir $(Build.BinariesDirectory)\tests
        cd $(Build.BinariesDirectory)\tests
        7z x $(Build.SourcesDirectory)\ci\helloworld.7z
        cd ..
        qmake $(Build.BinariesDirectory)\tests\helloworld
        mingw32-make
      }
    condition: |
      and(
        eq(variables['Agent.OS'], 'Windows_NT'),
        eq(variables['MODULE'], ''),
        eq(variables['SUBCOMMAND'], 'install-qt')
      )
    displayName: build test with qmake w/o extra module
```

**Comment by jdpurcell** [comment]
> 
> Could you look `ci/steps.yml` that defines steps to execute? There are MSVC related code in line 275.
> 

Nice hint, I see on line 240 where this `if` block doesn't handle msvc2022

```pwsh
      if ('$(ARCH)' -like '*msvc2019*') {
        Write-Host '##vso[task.setvariable variable=VSVER]2019'
      } elseif ('$(ARCH)' -like '*msvc2017*') {
        Write-Host '##vso[task.setvariable variable=VSVER]2017'
      } else {
        Write-Host '##vso[task.setvariable variable=VSVER]2015'
      }
```

jdpurcell:
Hmm so looking like the Windows jobs might need to be split up between [vmImage](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software) `windows-2019` and `windows-2022` depending on Qt version 🤔

jdpurcell:
Unless it's okay to run them all on `windows-2022`, I think it will work, but I don't know if it's acceptable to build with MSVC 2022 using Qt MSVC 2019 libraries.

jdpurcell:
Might only be a matter of time before they all need to be on `windows-2022` anyway, Windows runner images are only provided for the 2 latest versions per their policy, so `windows-2019` will be retired several months (presumably) after the `windows-2025` runner is released. I'd guess `windows-2019` will go away mid-2025.

tsteven4:
> I don't know if it's acceptable to build with MSVC 2022 using Qt MSVC 2019 libraries.

That should be fine.  We do it regularly on [gpsbabel.](https://github.com/GPSBabel/gpsbabel/blob/master/.github/workflows/windows.yml)

See https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170

> We've changed this behavior in Visual Studio 2015 and later versions. The runtime libraries and apps compiled by any of these versions of the compiler are binary-compatible. It's reflected in the C++ toolset major number, which starts with 14 for all versions since Visual Studio 2015. (The toolset version is v140 for Visual Studio 2015, v141 for 2017, v142 for 2019, and v143 for 2022). Say you have third-party libraries built by Visual Studio 2015. You can still use them in an application built by Visual Studio 2017, 2019, or 2022. There's no need to recompile with a matching toolset. The latest version of the Microsoft Visual C++ Redistributable package (the Redistributable) works for all of them.

ci/steps.yml
```diff
@@ -285,7 +290,13 @@ steps:
         cd $(Build.BinariesDirectory)\tests
         7z x $(Build.SourcesDirectory)\ci\helloworld.7z
         cd ..
-        qmake $(Build.BinariesDirectory)\tests\helloworld
+        
+        # Add QMAKE_CXXFLAGS for MSVC 2022
```
tsteven4, line 294: As a separate PR should we transition from qmake to cmake?  gpsbabel maintenance certainly got much easier when we did so.

jdpurcell:
I think there's only one job using win64_msvc2019 anyway, the one that tests a 6.5.3 Win ARM64 cross-compile.

@Kidev I tried [this](https://github.com/jdpurcell/aqtinstall/commit/8b259d76e946dfaeac3c9136812c870dc5c8ca05) through my Azure pipeline and it fixed the 6.8.1 Windows desktop job at least. The 6.8.1 wasm ones still fail but that isn't MSVC-related.

Kidev:
> I think there's only one job using win64_msvc2019 anyway, the one that tests a 6.5.3 Win ARM64 cross-compile.
> 
> @Kidev I tried [this](https://github.com/jdpurcell/aqtinstall/commit/8b259d76e946dfaeac3c9136812c870dc5c8ca05) through my Azure pipeline and it fixed the 6.8.1 Windows desktop job at least. The 6.8.1 wasm ones still fail but that isn't MSVC-related.

Thanks! I used a different take but if it fails I'll try yours. I'll setup a Azure account too to learn more about it, it'll always be a nice additional skill to have anyway

jdpurcell:
> I'll setup a Azure account too to learn more about it

I'm not even sure how I did it, I setup an Azure account for Pipelines a few years ago before a project I was working on moved to GitHub Actions. If I remember correctly, it was kind of confusing as to how the free access was supposed to work. Pretty sure I had to request a free grant as described [here](https://learn.microsoft.com/en-us/azure/devops/release-notes/2021/sprint-184-update) but they aren't taking new requests at this time.

Kidev:
[I managed to do it](https://github.com/Kidev/aqtinstall), Azure and Github are quite well integrated it's nice. I just had to change the azure-pipelines.yml to not use concurrency (the matrix thing):
```
trigger: [master]
pr: [master]

jobs:
- job: MacOS_Qt681
  displayName: 'MacOS - Qt 6.8.1'
  pool:
    vmImage: 'macOS-14'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-qt'
    QT_VERSION: '6.8.1'
    HOST: 'mac'
    TARGET: 'desktop'
    ARCH: 'clang_64'
    ARCHDIR: 'macos'
    QT_BINDIR: '$(Build.BinariesDirectory)/Qt/6.8.1/macos/bin'
  steps:
    - script: brew install p7zip
      displayName: Install 7zip
    - template: ci/steps.yml

- job: MacOS_Qt620_Charts
  displayName: 'MacOS - Qt 6.2.0 with Charts'
  dependsOn: MacOS_Qt681
  pool:
    vmImage: 'macOS-14'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-qt'
    QT_VERSION: '6.2.0'
    HOST: 'mac'
    TARGET: 'desktop'
    ARCH: 'clang_64'
    ARCHDIR: 'macos'
    MODULE: 'qtcharts qtnetworkauth'
    QT_BINDIR: '$(Build.BinariesDirectory)/Qt/6.2.0/macos/bin'
  steps:
    - script: brew install p7zip
      displayName: Install 7zip
    - template: ci/steps.yml

- job: Windows_Qt681_MSVC2022
  displayName: 'Windows - Qt 6.8.1 MSVC 2022'
  dependsOn: MacOS_Qt620_Charts
  pool:
    vmImage: 'windows-2019'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-qt'
    QT_VERSION: '6.8.1'
    HOST: 'windows'
    TARGET: 'desktop'
    ARCH: 'win64_msvc2022_64'
    ARCHDIR: 'msvc2022_64'
    MODULES_FOLDER: '$(System.DefaultWorkingDirectory)\CachedPowershellModules'
    startYear: $[format('{0:yyyy}', pipeline.startTime)]
    startMonth: $[format('{0:MM}', pipeline.startTime)]
    WIN_QT_BINDIR: '$(Build.BinariesDirectory)\Qt\6.8.1\msvc2022_64\bin'
  steps:
    - template: ci/steps.yml

- job: Linux_Qt681
  displayName: 'Linux - Qt 6.8.1'
  dependsOn: Windows_Qt681_MSVC2022
  pool:
    vmImage: 'ubuntu-22.04'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-qt'
    QT_VERSION: '6.8.1'
    HOST: 'linux'
    TARGET: 'desktop'
    ARCH: 'linux_gcc_64'
    ARCHDIR: 'gcc_64'
    QT_BINDIR: '$(Build.BinariesDirectory)/Qt/6.8.1/gcc_64/bin'
  steps:
    - template: ci/steps.yml

# Tools installation job
- job: Linux_Tools
  displayName: 'Linux - Install Tools'
  dependsOn: Linux_Qt681
  pool:
    vmImage: 'ubuntu-22.04'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-tool'
    HOST: 'linux'
    TARGET: 'desktop'
    TOOL1_ARGS: 'tools_qtcreator qt.tools.qtcreator'
    TOOL2_ARGS: 'tools_ifw'
    LIST_TOOL1_CMD: 'ls ./Tools/QtCreator/bin/'
    TEST_TOOL1_CMD: './Tools/QtCreator/bin/qbs --version'
    LIST_TOOL2_CMD: 'ls ./Tools/QtInstallerFramework/*/bin/'
    TEST_TOOL2_CMD: './Tools/QtInstallerFramework/*/bin/archivegen --version'
  steps:
    - template: ci/steps.yml

# WebAssembly build
- job: Linux_Qt673_WASM
  displayName: 'Linux - Qt 6.7.3 WebAssembly'
  dependsOn: Linux_Tools
  pool:
    vmImage: 'ubuntu-22.04'
  variables:
    PYTHON_VERSION: '3.12'
    SUBCOMMAND: 'install-qt'
    QT_VERSION: '6.7.3'
    HOST: 'all_os'
    TARGET: 'wasm'
    ARCH: 'wasm_multithread'
    ARCHDIR: 'wasm_multithread'
    IS_AUTODESKTOP: 'True'
    QT_BINDIR: '$(Build.BinariesDirectory)/Qt/6.7.3/wasm_multithread/bin'
    EMSDK_VERSION: '3.1.50'
    EMSDK_TAG: 'main'
  steps:
    - template: ci/steps.yml
```

ci/steps.yml
Kidev: It's a possibility, but I usually consider it like this:
- CMake is more generalist, and is the one I use by default because I'm used to it and it plays nice with additional libraries
- QMake is the faster and easier way to get it to work

Since it's just about installing Qt to test, I would say QMake is a better choice to not get tangled in CMake when we just want to test that our installs build

tsteven4: Most of these examples came from Qt, and they have converted them to cmake (adding CMakeLists.txt and deleting the *.pro files).  But I would suggest this is outside the scope of your PR anyway.

Kidev:
Updated the PR description to match all the changes. It should be good now, or very close to it. Just needs the workflow changes approved to check properly.

aqt/metadata.py
```diff
@@ -185,22 +185,58 @@ def get_semantic_version(qt_ver: str, is_preview: bool) -> Optional[Version]:
...
     elif len(qt_ver) >= 4:
-        return Version(major=int(qt_ver[:1]), minor=int(qt_ver[1:3]), patch=int(qt_ver[3:]))
+        return Version(major=int(qt_ver[0]), minor=int(qt_ver[1:3]), patch=int(qt_ver[3:]))
     elif len(qt_ver) == 3:
-        return Version(major=int(qt_ver[:1]), minor=int(qt_ver[1:2]), patch=int(qt_ver[2:]))
+        return Version(major=int(qt_ver[0]), minor=int(qt_ver[1]), patch=int(qt_ver[2]))
     elif len(qt_ver) == 2:
-        return Version(major=int(qt_ver[:1]), minor=int(qt_ver[1:2]), patch=0)
+        return Version(major=int(qt_ver[0]), minor=int(qt_ver[1]), patch=0)
+
```
Kidev, line 239: The changes of lines 234-238 are purely clarity related. They are equivalent but easier to read

Kidev:
Updated the PR description to match all the changes @miurahr. It should be good now, or very close to it. Just needs the workflow changes approved to check properly.

aqt/metadata.py
```diff
@@ -276,25 +320,33 @@ def to_url(self) -> str:
...
+                base = f"qt{version.major}_{qt_version_no_dots}"
+                return f"{base}/{base}"
+        elif version >= Version("6.5.0") and self.target == "wasm":
+            # Qt 6.5-6.7 WASM uses direct wasm_[single|multi]thread folder
+            if extension:
+                return f"qt{version.major}_{qt_version_no_dots}_{extension}"
+            return f"qt{version.major}_{qt_version_no_dots}"
+        # Pre-6.8 structure for non-WASM or pre-6.5 structure
+        return "{category}{major}_{ver}{ext}".format(
+            category=self.category,
```
tsteven4, line 339: The category handling is a bit different for >= 6.8.0 non-wasm, category is now assumed to be qt in that case.  category is either tools or qt.  Is there any case where the directory component is tools{major}_{ver}{ext}, either before or after 6.8.0?

Kidev: This return is there for the legacy versions. For Qt 6.8 non-WASM, this is returned:
```python
base = f"qt{version.major}_{qt_version_no_dots}"
return f"{base}/{base}"
```

All tests pass with this `to_folder`, older and newer versions of Qt, WASM of not

jdpurcell:
@Kidev I still expect the build will fail if you didn't update `vmImage`. Keep in mind these are pre-built VM images from Microsoft with a specific list of tools and versions pre-installed. The `windows-2019` image only comes with Visual Studio 2019 and the `windows-2022` image only comes with Visual Studio 2022, as [noted here](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software), and it's not feasible as far as I'm aware to install a different version.
image

As for the Windows Qt 6.8.1 WASM build failures which weren't related to Visual Studio, I found the problem was due to an incorrect `HostLibraryExecutables` setting in target_qt.conf. The patching wasn't working correctly on either Qt install (WASM nor the host); you can take a look at the Windows Qt 6.5.0 WASM build log under the `Run Aqt` step to see how many files are supposed to get patched, most of which wasn't happening for 6.8.1.

I fixed the Azure builds in these commits: [windows-2022](https://github.com/jdpurcell/aqtinstall/commit/1de5c759d84e34d8a0c23d81b313c13499aa45e4), [patching](https://github.com/jdpurcell/aqtinstall/commit/237947dc7f842ce9b18577cec69899e4f1ba167c). They were based off your 3f108bc8d698e725cdaa4f951bd200769af602b1 commit, i.e. I didn't need anything in 1d872244a8da9d98dd229fd847cfdc4769959b91, 254d66a8b04e572ee53bbab8b623f3eae8fb4161, 0c7abad791b91287571f2e75999d45a2ae28b91b, or bb33a225ef6f73bd10cc89e1897ae53a99aad523 to get all the Azure jobs working (I don't know if those commits were meant strictly to address the build failures or were partially for other reasons).

aqt/metadata.py
```diff
@@ -276,25 +320,33 @@ def to_url(self) -> str:
...
-            )
+            if self.target == "wasm":
+                # Qt 6.8+ WASM uses a split folder structure
+                folder = f"qt{version.major}_{qt_version_no_dots}"
+                if extension:
+                    folder = f"{folder}/{folder}_{extension}"
+                return folder
+            else:
+                base = f"qt{version.major}_{qt_version_no_dots}"
+                return f"{base}/{base}"
```
tsteven4, line 331: This is different than previous code.  Your code
1. doesn't use category
2. doesn't use extension
I can't find any instance in the sdk repo where something other than "qt" would work for category.  I am not as confident about extension, but I can't cite an example of this causing a failure.
In any event this bit will have to change to get src/doc/examples working (as in #857)

Kidev:
@jdpurcell Reverted to #[3f108bc8](https://github.com/miurahr/aqtinstall/commit/3f108bc8d698e725cdaa4f951bd200769af602b1), patched and signed off your commits @jdpurcell, and cherry picked back the fixes and tests to `get_semantic_version`

jdpurcell:
Of course one of the jobs we were most interested in fails again due to the bad mirror 😅

Kidev:
And twice in a row :')

jdpurcell:
At least it was a different job this time so we know they all work. I see the mirror in the blacklist of `ci/settings.ini` already, but it starts with `http://` instead of `https://`, I guess that's why it's still being used?

Kidev:
> At least it was a different job this time so we know they all work. I see the mirror in the blacklist of `ci/settings.ini` already, but it starts with `http://` instead of `https://`, I guess that's why it's still being used?

Oh thank god, I quickly checked and it started the same way I was worried! Indeed it's good then, but let's remove that mirror it's always this one

Kidev:
```console
INFO: pip is looking at multiple versions of aqtinstall to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement py7zr>=1.0.0-rc1 (from aqtinstall) (from versions: none)
ERROR: No matching distribution found for py7zr>=1.0.0-rc1
```
🫠

jdpurcell:
Doesn't it seem like only the Windows builds are encountering these download errors? It's not entirely a new issue because we've been having these random download issues (only with Windows too?) on other PRs, but I'm wondering now if it's somehow gotten worse on Windows Server 2022. If it keeps failing we could try moving [back to windows-2019](https://github.com/jdpurcell/aqtinstall/commit/2fc8037d08238080074c9430be996a383c6cd614) and see if it's at least not quite as error-prone.

jdpurcell:
Just another random thought, could it be something like Azure has runner instances for the different OS types in different data centers / IP ranges so the Windows ones are more likely to get redirected to these random mirrors by Qt's CDN (e.g. if it works by geolocation)? If CI tried `qt.mirror.constant.com` (that's what the Mac/Linux ones typically use) as its preferred mirror I wonder if it would help.

Kidev:
@jdpurcell Indeed but I'm much more enclined to bet on some caching there. All the other builds passed earlier so their cache is kept, which is not the case for the builds that failed up until recently. I can perfectly build those on Linux, Win11 and macOS VM on my PC. I'm not going to try to fix something that is not broken honestly... Plus the longer this goes the more the goalposts move, and I'm not going to fix conflicts forever until it gets merged or some more investigation is done. CI is annoying but CI you don't manage/admin is even worse

Kidev:
Thanks @jdpurcell

aqt/settings.ini
```diff
@@ -17,7 +17,7 @@ retry_backoff: 0.1
 max_retries_on_checksum_error: 5
 max_retries_to_retrieve_hash: 5
 hash_algorithm: sha256
-INSECURE_NOT_FOR_PRODUCTION_ignore_hash: False
+INSECURE_NOT_FOR_PRODUCTION_ignore_hash: True
```
jdpurcell, line 20: @Kidev You accidentally changed this in 7a2a41b66d9cdecacdcd812e738b4117e1356dc5. Also maybe worth cleaning up the commits from earlier so it doesn't clutter the history. Let me show you something, maybe it's obvious, but I didn't know GitHub Desktop had a "Reorder Commit" feature until just recently. I find it quite powerful because it's kind of an alternate way to interactive rebase, when used in combination with squash. To be honest I've never figured out how to do a real interactive rebase, I used to get so annoyed with it I would just hard reset back to some point and then cherry-pick back in and amend stuff as I go 🤣. I was amazed when I figured out the reorder/squash trick.

https://github.com/user-attachments/assets/d29d4f7d-7aa6-4257-96c2-550609a34c3e

Kidev: I usually don't squash in PR because I either squash merge, so it's later and the branch keeps all history, or I merge without rebase, so the history is not cluttered either. I usually reset when I want to trash the latest commits quickly and I'm alone on the branch

I do everything using the CLI I'm too used to it. `git rebase -i` is much easier than it looks:
output

You even have all the commands to use when you enter the interactive mode:
```console
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup [-C | -c] <commit> = like "squash" but keep only the previous
#                    commit's log message, unless -C is used, in which case
#                    keep only this commit's message; -c is same as -C but
#                    opens the editor
# x, exec <command> = run command (the rest of the line) using shell
# b, break = stop here (continue rebase later with 'git rebase --continue')
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
#         create a merge commit using the original merge commit's
#         message (or the oneline, if no original merge commit was
#         specified); use -c <commit> to reword the commit message
# u, update-ref <ref> = track a placeholder for the <ref> to be updated
#                       to this position in the new commits. The <ref> is
#                       updated at the end of the rebase
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#```

**Comment by jdpurcell** [comment]
Regarding the current error, I would propose https://github.com/jdpurcell/aqtinstall/commit/a6b8a875dd047be9510db888f7f90d602cfa4f42. `downloadBinaryFile` already tries to catch `requests.exceptions.Timeout` which would cover both `ConnectTimeout` and `ReadTimeout` errors, but due to `stream=True` I believe the initial exception handling block only applies to the initial part of the connection (connect / receive headers?). This change would look for `ReadTimeout` during the streaming part of the download and raise an `ArchiveConnectionError` which should allow retries to occur.

**Comment by Kidev** [comment]
> [!WARNING]
> Found bug not checked by either tests or CI

When installing for WASM using `--autodesktop` to automatically download Qt for the host, the host does not download the same modules as the target

**Comment by Kidev** [comment]
In order to make the testing of Qt WASM for Qt6.8+ possible, I'm changing the working of `--autodesktop`. Instead of whatever it was doing before, now it simply detects when it is summoned, and initiates another run of `run_install_qt`. It basically acts the way a client would act if they wanted to install WASM and the host. The second run copies the modules, extensions, version, and just replaces `all_os` by the relevant host (`linux_gcc_64`, `mingw`...) and replaces the target (`wasm`) by `desktop`.

**Comment by Kidev** [comment]
Hm I think it's due to some CRLF and the regex that is used that causes issues. The error is just a slight mismatch between the expected output and the one received, not an error. I'm fixing it though

**Comment by tsteven4** [comment]
The failing wasm test (FAILED tests/test_install.py::test_install_qt6_wasm_autodesktop[6.8.0-680-wasm_singlethread]) from https://github.com/miurahr/aqtinstall/actions/runs/12447000072/job/34756107879
seems to be a result of interleaving of Finished and Downloading log entries.  Also there are multiple "Time elapsed" entries in the log, but only one in the pattern.

I think the pattern is https://github.com/miurahr/aqtinstall/blob/84fad94b6ff892815e0e84532a2d6a7119bb226a/tests/test_install.py#L1534-L1549

Replacing the newlines in the raw log I get

> 2024-12-22T01:04:08.6356821Z  +  where None = <built-in method match of re.Pattern object at 0x55c22f72a160>('INFO    : aqtinstall(aqt) v3.1.22.dev22 on Python 3.12.8 [CPython GCC 11.4.0]
INFO    : You are installing the Qt6-WASM version of Qt, which requires that the desktop version of Qt is also installed. Now installing Qt: desktop 6.8.0 linux_gcc_64
INFO    : aqtinstall(aqt) v3.1.22.dev22 on Python 3.12.8 [CPython GCC 11.4.0]
INFO    : Found extension qtwebengine
INFO    : Found extension qtpdf
INFO    : Downloading qtcharts...
Finished installation of qtcharts-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04424284
INFO    : Downloading qtquick3d...
Finished installation of qtquick3d-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04443704
INFO    : Downloading qtbase...
Finished installation of qtbase-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04172059
INFO    : Downloading qtsvg...
Finished installation of qtsvg-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04188181
INFO    : Downloading qtdeclarative...
Finished installation of qtdeclarative-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04231454
INFO    : Downloading qttools...
Finished installation of qttools-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.03957928
INFO    : Downloading qttranslations...
Finished installation of qttranslations-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.04014839
INFO    : Downloading qtwayland...
Finished installation of qtwayland-Linux-RHEL_8_8-GCC-Linux-RHEL_8_8-X86_64.7z in 0.03933383
INFO    : Downloading icu...
Finished installation of icu-linux-Rhel8.6-x86_64.7z in 0.04046533
INFO    : Finished installation
INFO    : Time elapsed: 0.51118661 second
INFO    : Found extension qtwebengine
INFO    : Found extension qtpdf
INFO    : Downloading qtcharts...
Finished installation of qtcharts-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.03493405
INFO    : Downloading qtquick3d...
Finished installation of qtquick3d-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.03722132
INFO    : Downloading qtbase...
Finished installation of qtbase-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.04040757
INFO    : Downloading qtsvg...
Finished installation of qtsvg-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.04199367
INFO    : Downloading qtdeclarative...
Finished installation of qtdeclarative-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.04036353
INFO    : Downloading qttools...
Finished installation of qttools-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.04069645
INFO    : Downloading qttranslations...
Finished installation of qttranslations-Windows-Windows_10_22H2-Clang-Windows-WebAssembly-X86_64.7z in 0.04391751
INFO    : Patching /tmp/tmpsn5is4dp/6.8.0/wasm_singlethread/bin/qmake
INFO    : Patching /tmp/tmpsn5is4dp/6.8.0/wasm_singlethread/bin/qtpaths
INFO    : Patching /tmp/tmpsn5is4dp/6.8.0/wasm_singlethread/bin/qmake6
INFO    : Patching /tmp/tmpsn5is4dp/6.8.0/wasm_singlethread/bin/qtpaths6
INFO    : Finished installation
INFO    : Time elapsed: 0.89748959 second
')

**Comment by tsteven4** [comment]
The first failure is trying to match the following.  The pattern expects only one "Time elasped" entry.

> INFO    : aqtinstall(aqt) v3.1.22.dev22 on Python 3.12.8 [CPython GCC 11.4.0]
INFO    : You are installing the MSVC Arm64 version of Qt, which requires that the desktop version of Qt is also installed. Now installing Qt: desktop 6.5.2 win64_msvc2019_64
INFO    : aqtinstall(aqt) v3.1.22.dev22 on Python 3.12.8 [CPython GCC 11.4.0]
INFO    : Downloading qtbase...
Finished installation of qtbase-windows-win64_msvc2019_64.7z in 0.03976085
INFO    : Finished installation
INFO    : Time elapsed: 0.05573336 second
INFO    : Downloading qtbase...
Finished installation of qtbase-windows-win64_msvc2019_arm64.7z in 0.04444390
INFO    : Patching /tmp/tmpclrio675/6.5.2/msvc2019_arm64/bin/qmake.bat
INFO    : Patching /tmp/tmpclrio675/6.5.2/msvc2019_arm64/bin/qtpaths.bat
INFO    : Patching /tmp/tmpclrio675/6.5.2/msvc2019_arm64/bin/qmake6.bat
INFO    : Patching /tmp/tmpclrio675/6.5.2/msvc2019_arm64/bin/qtpaths6.bat
INFO    : Finished installation
INFO    : Time elapsed: 0.12507072 second

**Comment by Kidev** [comment]
There was a mistake in the autodesktop logic. It's all fixed, and the output log while installing with `--autodesktop` is much more clear about what it is doing

**Review by Kidev — COMMENTED** [review]
Requested changes made, added requested tests, all checks and tests pass

**Comment by vkuznetsovgn** [comment]
Hi @Kidev .

I'm currently testing your PR. Everything is going great except I cannot figure out how to install WASM now for Qt 6.6 and lower?

When I'm running the command from the doc `aqt list-qt linux desktop --arch 6.6.3` it gives me only `gcc_64`.
However, the same command from the main branch gives me `gcc_64 wasm_singlethread wasm_multithread`.

**Comment by Kidev** [comment]
> Hi @Kidev .
> 
> 
> 
> Thanks for your work!
> 
> 
> 
> I'm currently testing your PR. Everything is going great except I cannot figure out how to install WASM now for Qt 6.6 and lower?
> 
> 
> 
> When I'm running the command from the doc `aqt list-qt linux desktop --arch 6.6.3` it gives me only `gcc_64`.
> 
> However, the same command from the main branch gives me `gcc_64 wasm_singlethread wasm_multithread`.

Hm weird, nice catch, I'm going to check and fix that this week probably. It should still work to install it the old way though since there are tests passing on old wasm version iirc

**Comment by miurahr** [comment]
Dear @Kidev and @vkuznetsovgn 
Thank you for the hard work and the review efforts.

The code quality looks good for merge, and there is also good coverage.
 
> I'm currently testing your PR. Everything is going great except I cannot figure out how to install WASM now for Qt 6.6 and lower?
> 
> When I'm running the command from the doc `aqt list-qt linux desktop --arch 6.6.3` it gives me only `gcc_64`. However, the same command from the main branch gives me `gcc_64 wasm_singlethread wasm_multithread`.

Actually there are three WASM related folders on download.qt.io site.
You can observe folders on https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/

- qt6_663
- qt6_663_wasm_singlethread
- qt6_663_wasm_multithread

It is why main branch report three. 

I think it is not a problem the WASM branch, here, report only `gcc_64`  but a command like `$ aqt list-qt all_os wasm --arch 6.6.3` reports these two WASM related names as same as version 6.7.3 @Kidev suggested, because these two are valuable only for a WASM installation.

File: ci/settings.ini [hunk 10]
```diff
@@ -2,7 +2,7 @@
 
 [aqt]
 concurrency: 2
-baseurl: https://download.qt.io
+baseurl: https://qt.mirror.constant.com
```
miurahr, line 5: I don't like to change it into mirror site.
It may be works well and only for CI process, but it is not encouraged for users.

It is better to add comment in the file why it uses mirror and the concern.

vkuznetsovgn:
Hi @Kidev .

`aqt install-qt linux desktop 6.6.3 wasm_multithread --autodesktop` indeed works.

Hi @miurahr.

`aqt list-qt all_os wasm --arch 6.6.3` outputs

```
ERROR   : No architectures available for this request.
```

I believe it's something with list-qt command only then.

Kidev:
There is indeed an issue with versions 6.5.x and 6.6.x not reporting wasm_* in desktop while fetching archs. I located the issue and I'm fixing it. It is indeed just an issue on the list-qt. About the fact that `aqt list-qt all_os wasm --arch 6.6.3` outputs an error, this is expected, as the "all_os" OS target exists only for versions 6.7.x+

ci/settings.ini
Kidev: I'm fixing this in the commit that also fixes the list-qt issue!

Kidev (commented):
Fixed `list-qt` issues for WASM on versions 6.5.x and 6.6.x, and added comment for the CI settings to make it clear why it is a third party mirror that is used

miurahr (approved):

miurahr:
I would like to merge it as squashed commit.

miurahr:
Great appreciation for the hard work @Kidev !!

Kidev:
> Great appreciation for the hard work @Kidev !!

Happy to help on such a nice and useful tool ! :)

PR #853, opened 2024-12-14
Patch additional Qt scripts

jdpurcell:
Recently, I wanted to use `qtpaths`, because when cross-compiling for ARM64 Windows for example, one must invoke the host Qt's `windeployqt` and pass in either the `--qmake` or `--qtpaths` argument in order for `wideployqt` to find the target Qt's configuration. `--qmake` works, but it has been marked deprecated for a while now and writes a warning to the console recommending to use `--qtpaths` instead. `aqt` only patches the `qmake` script though, leaving an incorrect path in `qtpaths`. This PR corrects that, as well as patching `qmake6` and `qtpaths6` for completeness.

It is limited to the appropriate versions, since `qtpaths` was introduced with [QTBUG-97821](https://bugreports.qt.io/browse/QTBUG-97821), and the versioned variants were introduced with [QTBUG-109024](https://bugreports.qt.io/browse/QTBUG-109024). Note that although the former was marked as fixed in 6.3, it was also cherry-picked back to the 6.2 branch. I checked the version tags in Qt's repo to find exactly when each change was added, and download zip files of the published binaries from their mirrors to make sure the version logic is correct.

I [ran a test](https://github.com/jdpurcell/QtHelloWorld/actions/runs/12333456922) via GitHub Actions using my modified `aqt` and can confirm it works properly for many Qt versions, allowing one to cross-compile for ARM64 Windows using the `qtpaths` argument when possible to avoid the deprecation warning.

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -184,13 +184,22 @@ class Updater:
                 newpath=bytes(str(self.prefix), "UTF-8"),
             )
 
-    def patch_qmake_script(self, base_dir, qt_version: str, os_name: str, desktop_arch_dir: str):
+    def patch_qt_scripts(self, base_dir, version_dir: str, os_name: str, desktop_arch_dir: str, version: Version):
         sep = "\\" if os_name == "windows" else "/"
-        patched = sep.join([base_dir, qt_version, desktop_arch_dir, "bin"])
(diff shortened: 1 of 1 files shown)


jdpurcell:
Current failures are due to mirror timeouts; will retry later.

miurahr:
LGTM.

PR #864, opened 2024-12-19
appease mypy Version no-untyped-call

tsteven4:
Without the type hint we get:
error: Call to untyped function "Version" in typed context  [no-untyped-call]

Note that our Version is derived from semantic_version, which doesn't have type hints.

### DIFF
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -79,7 +79,7 @@ class Version(SemanticVersion):
         prerelease=None,
         build=None,
         partial=False,
-    ):
+    ) -> None:
         if version_string is None:
             super(Version, self).__init__(
                 version_string=None,


tsteven4:
I also note related mypy errors:
```
aqt/metadata.py:67:15: error: Class cannot subclass "SemanticVersion" (has type
"Any")  [misc]
    class Version(SemanticVersion):
                  ^
aqt/metadata.py:73:5: error: Function is missing a type annotation for one or
more arguments  [no-untyped-def]
        def __init__(
        ^
aqt/metadata.py:108:5: error: Function is missing a type annotation
[no-untyped-def]
        def __str__(self):
        ^
aqt/metadata.py:114:5: error: Function is missing a return type annotation
[no-untyped-def]
        def permissive(cls, version_string: str):
        ^
```

miurahr:
Good catch.

Issue #867, opened 2024-12-23
[Bug]: Failure when installing example on Windows


vserdyuk:
### Bug description

Command

```
aqt install-example --outputdir D:\Qt2 windows 6.8.0
```

fails to complete on Windows 11 with aqt v3.1.21. Was working fine with v3.1.19.

I had some modifications in settings.ini:

- different baseurl: https://ftp.fau.de/qtproject/
- INSECURE_NOT_FOR_PRODUCTION_ignore_hash: True
- https://download.qt.io and https://master.qt.io/ moved to blacklist

```
[DEFAULTS]

[aqt]
concurrency: 4
baseurl: https://ftp.fau.de/qtproject/
7zcmd: 7z
print_stacktrace_on_error: False
always_keep_archives: False
archive_download_location: .
min_module_size: 41

[requests]
connection_timeout: 3.5
response_timeout: 30
max_retries_on_connection_error: 5
retry_backoff: 0.1
max_retries_on_checksum_error: 5
max_retries_to_retrieve_hash: 5
hash_algorithm: sha1
INSECURE_NOT_FOR_PRODUCTION_ignore_hash: True

[mirrors]
trusted_mirrors:
    https://ftp.fau.de/qtproject/
blacklist:
    https://download.qt.io
    http://mirrors.ocf.berkeley.edu
    http://mirrors.tuna.tsinghua.edu.cn
    http://mirrors.geekpie.club
    https://master.qt.io/
fallbacks:
    https://qtproject.mirror.liquidtelecom.com/
    https://mirrors.aliyun.com/qt/
    https://mirrors.ustc.edu.cn/qtproject/
    https://ftp.jaist.ac.jp/pub/qtproject/
    https://ftp.yz.yamagata-u.ac.jp/pub/qtproject/
    https://qt-mirror.dannhauer.de/
    https://ftp.fau.de/qtproject/
    https://mirror.netcologne.de/qtproject/
    https://mirrors.dotsrc.org/qtproject/
    https://www.nic.funet.fi/pub/mirrors/download.qt-project.org/
    https://mirrors.ukfast.co.uk/sites/qt.io/
    https://ftp2.nluug.nl/languages/qt/
    https://ftp1.nluug.nl/languages/qt/
    https://qt.mirror.constant.com/
```

### Expected behavior

Successful download of examples

### aqt and python version

aqtinstall(aqt) v3.1.21 on Python 3.12.6 [CPython MSC v.1940 64 bit (AMD64)]

[...]

PR #868, opened 2024-12-24
Fix: variable become undefined when extension not exist or connection…

miurahr:
… error

- Fixed the issue #867

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -442,11 +442,12 @@ class QtArchives:
                 try:
                     extensions_xml_text = self._download_update_xml(extensions_xml_url, True)
                 except ArchiveDownloadError:
-                    # In case _download_update_xml ignores the hash and tries to get the url.
+                    # In case _download_update_xml failed to get the url because of no extension.
                     pass
-                if extensions_xml_text:
-                    self.logger.info("Found extension {}".format(ext))
-                    update_xmls.append(UpdateXmls(extensions_target_folder, extensions_xml_text))
+                else:
+                    if extensions_xml_text:
+                        self.logger.info("Found extension {}".format(ext))
+                        update_xmls.append(UpdateXmls(extensions_target_folder, extensions_xml_text))
 
         self._parse_update_xmls(update_xmls, target_packages)
 


miurahr:
@tsteven4  Please check the change?

aqt/helper.py
```diff
@@ -356,20 +358,21 @@ def __init__(self) -> None:
                     self.loggingconf = os.path.join(os.path.dirname(__file__), "logging.ini")
 
     def load_settings(self, file: Optional[Union[str, TextIO]] = None) -> None:
-        if file is not None:
-            if isinstance(file, str):
-                result = self.config.read(file)
-                if len(result) == 0:
-                    raise IOError("Fails to load specified config file {}".format(file))
-                self.configfile = file
+        if self.config is not None:
```
miurahr, line 361: It is better to be early return here.

tsteven4:
I agree I messed this up.  I have an alternative fix in process.  I will submit it for consideration.

aqt/archives.py
```diff
@@ -442,11 +442,12 @@ def _get_archives_base(self, name, target_packages):
                 try:
                     extensions_xml_text = self._download_update_xml(extensions_xml_url, True)
                 except ArchiveDownloadError:
-                    # In case _download_update_xml ignores the hash and tries to get the url.
+                    # In case _download_update_xml failed to get the url because of no extension.
```
tsteven4, line 445: The primary failure when the extension doesn't exist is ChecksumDownloadFailure, which is caught by download_update_xml.  I had to modify the code to simulate getting through that and failing with an ArchiveDownloadError.

tsteven4: and I see in #867 the OP had set INSECURE_NOT_FOR_PRODUCTION_ignore_hash: True

So I think the original comment is more accurate.

aqt/archives.py
```diff
@@ -442,11 +442,12 @@ def _get_archives_base(self, name, target_packages):
...
                     pass
-                if extensions_xml_text:
-                    self.logger.info("Found extension {}".format(ext))
-                    update_xmls.append(UpdateXmls(extensions_target_folder, extensions_xml_text))
+                else:
+                    if extensions_xml_text:
+                        self.logger.info("Found extension {}".format(ext))
+                        update_xmls.append(UpdateXmls(extensions_target_folder, extensions_xml_text))
```
tsteven4, line 450: I moved this into the try block but putting them in the else block may be preferred.  There are two places in metadata.py
related to "except (ChecksumDownloadFailure, ArchiveDownloadError)" where I also added additional code to the try block where an else block might be preferred.

tsteven4 (commented):
While you are in here it would be good to change https://github.com/miurahr/aqtinstall/blob/98fda2ebaf68baf558101c3cf1c2ba00c69bfa14/aqt/archives.py#L436
to
`for ext in QtRepoProperty.known_extensions(self.version):`
so we only have the list of extensions in one place.

aqt/archives.py
miurahr: When `_download_update_xml` method failed to download sha256 checksum, it does not throw exception but set `xml_hash = None` and call `get_url`.
The exception is thrown from  `get_url` that try to `Updates.xml` itself.

So when got `ArchiveDownloadError` here, it means that it failed to download `Update.xml` not hash.

miurahr:
> While you are in here it would be good to change

It is out of scope of the fix.

aqt/archives.py
tsteven4: Assuming Settings.ignore_hash is False download_update_xml catches ChecksumDownloadFailure and RETURNS None.  It doesn't call get_url (silent = True).  This is the normal production behavior.

The OP set ignore_hash to True which behaves as you indicated.

tsteven4 (approved):

PR #828, opened 2024-10-06
restore CI pipeline test for list-* commands

ddalcino:
Restore CI pipeline test for list-* commands.

See https://github.com/miurahr/aqtinstall/commit/bdf8b4b1f99795d3733bee4b4d0eeaa629637af1#r147631586

I'm slightly concerned about using strings like `list-*` in Bash because of potential issues with wildcard character expansion. As long as these strings are quoted, I think they should be fine though.

PR #870, opened 2024-12-24
fix potential erroneous handling of extension data.

tsteven4:
If an ArchiveDownloadError was raised:
1. on the first pass through the loop an error could occur "UnboundLocalError:
   cannot access local variable 'extensions_xml_text' where it is not
   associated with a value"
2. on subsequent passes the previous value of extensions_xml_text could
   be reused inappropriately.

Also use QtRepoProperty.known_extensions for improved maintainability,
i.e. the list of possible extensions is only defined in one place.


tsteven4:
We need the list-qt test back online (#828)!

tsteven4:
I still working on the test.  Feel free to close this PR and go with #868 if you prefer.

PR #871, opened 2024-12-28
enhance test_install to exercise extensions

tsteven4:
This test can be used to reproduce #867 by changing INSECURE_NOT_FOR_PRODUCTION_ignore_hash to True in settings.ini.  It can also be used to verify #868 resolves the issue.


miurahr:
The test failure is caused by bad mirror site.
The change is good and merge it now.

PR #883, opened 2025-01-29
Commercial fixes, CI tests, more tests, coverage

Kidev:
- Fixes issues with clashing auth method in commercial calls
- Add a file dedicated to commercial testing
- Add CI tests
- Improve coverage

WIP

Note: I think my auto-format is not well configured for .sh files, I'll fix to not overwrite lines I did not touch for nothing

PR #891, opened 2025-02-15
Explicitly use python 3.13

xavier2k6:


xavier2k6:
@Kidev Will #883 address tox failures?

PR #895, opened 2025-02-28
chore(ci): avoid zipped resources from repository

miurahr:
For security reasons, I want to avoid 7z/zip archives from the repository.

- Expand all test sources on the repository
- Download jom.zip from the genuine repository and check md5sum

PR #896, opened 2025-03-01
chore(ci): improve msvc detections

miurahr:


ci/steps.yml
```diff
@@ -243,9 +243,15 @@ steps:
...
     }
-    if ('$(ARCH)' -like '*msvc*') {
+    if ('$(ARCH)' -like '*msvc2022*') {
       Write-Host '##vso[task.setvariable variable=VSVER]2022'
     }
+    if ('$(ARCH)' -like '*msvc2019*') {
+      Write-Host '##vso[task.setvariable variable=VSVER]2019'
+    }
+    if ('$(ARCH)' -like '"*_arm64') {
+      Write-Host '##vso[task.setvariable variable=ARCHITECTURE]arm64'
```
jdpurcell, line 253:
```suggestion
      Write-Host '##vso[task.setvariable variable=ARCHITECTURE]amd64_arm64'
```
You would only use `arm64` here if the host were ARM64 too, it's different [when cross-compiling](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#vcvarsall-syntax).

jdpurcell:
I think the Azure build got confused because of #895 getting merged after the pipeline run started. I.e. the matrices generator ran, then the merge happened, then when the other jobs ran and they checked out `refs/remotes/pull/896/merge`, the repo was in a different state. Hopefully it will work if you re-run it.

jdpurcell:
@miurahr I fixed everything in #897.

PR #898, opened 2025-03-02
Fix official installer on macOS

jdpurcell:
Fixes #889. In addition to the incorrect download URL, the file we get is a macOS `.dmg` (disk image) which cannot be executed directly, so this handles mounting/extracting the contents. After these changes, I was able to successfully complete a [macOS build](https://github.com/jdpurcell/QtHelloWorld/actions/runs/13610178522/job/38046297825) via GitHub Actions with the official installer.

PR #900, opened 2025-03-04
feat: install-qt-official checks hash for downloaded binary

miurahr:
The current `install-qt-officiial` feature uses requests library directory. 
This PR replace `requests.get` with `aqt.helper.downloadBinaryFile` 

Also, this updates functional test to adopt a method signature change, and to use of tmp_path fixture.

### DIFF
--- a/aqt/installer.py
+++ b/aqt/installer.py
@@ -63,6 +63,7 @@ from aqt.helper import (
     downloadBinaryFile,
     extract_auth,
     get_hash,
+    get_os_name,
     get_qt_installer_name,
     retry_on_bad_connection,
     retry_on_errors,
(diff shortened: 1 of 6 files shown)


miurahr:
The change here is discussed in #898.

jdpurcell:
@miurahr Do you think the download code in `install.py` for the `list-qt-commercial` command should also be updated?
https://github.com/miurahr/aqtinstall/blob/0ffe65700c50c8175a80b804e579df7b4ee99da3/aqt/installer.py#L907-L919

miurahr:
> @miurahr Do you think the download code in `install.py` for the `list-qt-commercial` command should also be updated?

Yes, I am working it for now.

aqt/installer.py
```diff
@@ -903,20 +904,8 @@ def run_list_qt_commercial(self, args) -> None:
...
-            response.raise_for_status()
-
-            with open(installer_path, "wb") as f:
-                for chunk in response.iter_content(chunk_size=8192):
-                    f.write(chunk)
-
-            if get_os_name() != "windows":
-                os.chmod(installer_path, 0o500)
+            timeout = (Settings.connection_timeout, Settings.response_timeout)
+            download_installer(Settings.baseurl, installer_filename, get_os_name(), temp_path, timeout)
```
jdpurcell, line 908:
```suggestion
            download_installer(Settings.baseurl, installer_filename, get_os_name(), installer_path, timeout)
```

PR #901, opened 2025-03-12
Fix missing log statement for target_qt.conf

jdpurcell:
I thought there was a problem with target_qt.conf patching for WASM recently because I didn't see it in the console output, but it was actually working fine! This function just wasn't writing to the logger like the other patching functions 😄


jdpurcell:
The tests should pass now.

PR #913, opened 2025-04-17
Fallback to mac-x64 when mac-arm64 installer is missing

alex-spataru:
macOS ARM (Apple Silicon) systems currently cause a KeyError in `get_qt_installer_name()` because no `mac-arm64` installer exists in the Qt online releases. This patch adds a fallback to `mac-x64`, which runs fine under Rosetta and is the only available macOS installer for the moment.

The patch also includes a runtime check to raise a clearer error message if no installer is found for the detected platform.

Please let me know if I need to change anything in my patch. Thanks!

### DIFF
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -147,13 +147,16 @@ def get_qt_installers() -> dict[str, str]:
 
 def get_qt_installer_name() -> str:
     installer_dict = get_qt_installers()
-    
-    # If mac-arm64 is missing but mac-x64 exists, use that
     os_arch = get_os_arch()
+    
     if os_arch == "mac-arm64" and os_arch not in installer_dict:
         fallback = "mac-x64"
         if fallback in installer_dict:
+            print("Warning: Falling back to mac-x64 installer for mac-arm64 platform.")
             return installer_dict[fallback]
+
+    if os_arch not in installer_dict:
+        raise RuntimeError(f"No installer found for platform: {os_arch}")
             
     return installer_dict[get_os_arch()]
 
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -147,6 +147,13 @@ def get_qt_installers() -> dict[str, str]:
 
 def get_qt_installer_name() -> str:
     installer_dict = get_qt_installers()
+    
+    # If mac-arm64 is missing but mac-x64 exists, use that
+    if os_arch == "mac-arm64" and os_arch not in installer_dict:
+        fallback = "mac-x64"
+        if fallback in installer_dict:
+            return installer_dict[fallback]
+            
     return installer_dict[get_os_arch()]
 
 
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -149,6 +149,7 @@ def get_qt_installer_name() -> str:
     installer_dict = get_qt_installers()
     
     # If mac-arm64 is missing but mac-x64 exists, use that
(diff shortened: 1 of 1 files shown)


Kidev:
Good idea for the more graceful handling of unknown os-arch.  
But Apple silicon is ARM anyway isn't it? So we should never have mac-arm64 in the installer name in the future. Trying to predict what the Qt team will do to their repo in the future is hard, but I think this may be better handle more upstream:

```python
def get_os_arch() -> str:
    """
    Returns a simplified os-arch string for the current system
    """
    os_name = get_os_name()

    machine = platform.machine().lower()
    if machine in ["x86_64", "amd64"]:
        arch = "x64"
    elif machine in ["arm64", "aarch64"] and os_name != "mac": # HERE
        arch = "arm64"
    else:
        arch = "x64"  # Default to x64 for mac and unknown archs

    return f"{os_name}-{arch}"
```
```python
def get_qt_installer_name() -> str:
    installer_dict = get_qt_installers()
    os_arch = get_os_arch()

    if not os_arch in installer_dict:
        raise RuntimeError(f"No installer found for platform: {os_arch}")

    return installer_dict[os_arch]
```

And adding `mac` to the list that allows for arch-less oses:
```python
def get_qt_installers() -> dict[str, str]:
...
            for os_type in os_types:
                if os_type.lower() in filename.lower():
                    # Found an OS match, now look for architecture
                    if "arm64" in filename.lower():
                        installers[f"{os_type}-arm64"] = filename
                    elif "x64" in filename.lower():
                        installers[f"{os_type}-x64"] = filename
                        installers[os_type] = filename
                    else:
                        # Handle case with no explicit architecture
                        installers[os_type] = filename

        return installers
...
```

Kidev (changes requested):
Good idea for the more graceful handling of unknown os-arch.  
But Apple silicon is ARM anyway isn't it? So we should never have mac-arm64 in the installer name in the future. Trying to predict what the Qt team will do to their repo in the future is hard, but I think this may be better handled more upstream:

```python
def get_os_arch() -> str:
    """
    Returns a simplified os-arch string for the current system
    """
    os_name = get_os_name()

    machine = platform.machine().lower()
    if machine in ["x86_64", "amd64"]:
        arch = "x64"
    elif machine in ["arm64", "aarch64"] and os_name != "mac": # HERE
        arch = "arm64"
    else:
        arch = "x64"  # Default to x64 for mac and unknown archs

    return f"{os_name}-{arch}"
```
```python
def get_qt_installer_name() -> str:
    installer_dict = get_qt_installers()
    os_arch = get_os_arch()

    if not os_arch in installer_dict:
        raise RuntimeError(f"No installer found for platform: {os_arch}")

    return installer_dict[os_arch]
```

And adding `mac` to the list that allows for arch-less oses:
```python
def get_qt_installers() -> dict[str, str]:
...
            for os_type in os_types:
                if os_type.lower() in filename.lower():
                    # Found an OS match, now look for architecture
                    if "arm64" in filename.lower():
                        installers[f"{os_type}-arm64"] = filename
                    elif "x64" in filename.lower():
                        installers[f"{os_type}-x64"] = filename
                        installers[os_type] = filename
                    else:
                        # Handle case with no explicit architecture
                        installers[os_type] = filename

        return installers
...
```

micheletimax:
Is there any update for the resolution of this patch? It does still seem not to be working

PR #914, opened 2025-04-19
update host windows_arm64 files as we do with windows.

tsteven4:
Qt 6.8 has added Windows on Arm as a supported configuration.  The restriction "Windows on ARM is only supported as a deployment target. Applications have to be cross-compiled from an x86-64 Windows machine and deployed to target." has been removed.

This PR simply treats windows and windows_arm64 hosts identically when patching files and creating scripts.  Some of these cases don't actually apply to WoA, but I treated them all identically for consistency.  For example patch_qt_core won't ever do anything on WoA because the windows_arm64 hosts wasn't ever supported with Qt 5.

Note other issues prevent aqt from running on WoA machines, e.g. https://codeberg.org/miurahr/pyppmd/issues/128.  However, this at least allows the WoA install to be done on another machine type and subsequently transferred to the WoA machine for use (with an identical path).

### DIFF
--- a/aqt/updater.py
+++ b/aqt/updater.py
@@ -185,11 +185,11 @@ class Updater:
             )
 
     def patch_qt_scripts(self, base_dir, version_dir: str, os_name: str, desktop_arch_dir: str, version: Version):
-        sep = "\\" if os_name == "windows" else "/"
+        sep = "\\" if os_name.startswith("windows") else "/"
         patched = sep.join([base_dir, version_dir, desktop_arch_dir, "bin"])
 
         def patch_script(script_name):
-            script_path = self.prefix / "bin" / (script_name + ".bat" if os_name == "windows" else script_name)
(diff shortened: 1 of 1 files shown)


jdpurcell:
In case anyone missed the announcement, the `windows-11-arm` hosted runner is [now in public preview](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/). So Windows ARM64 support would indeed be great! I encountered the `pyppmd` compilation issue to [as I was testing](https://github.com/jurplel/install-qt-action/pull/273#issuecomment-2803388199) WoA support for `install-qt-action`.

jdpurcell:
I found that it's possible to work around the `pyppmd` issue by [requesting x64 Python](https://github.com/jdpurcell/QtHelloWorld/commit/e6dea15afd2cf593eded93a19ab84669f30d80d6) from the `setup-python` action.

miurahr (approved):

PR #941, opened 2025-09-16
fix installation of extensions for windows arm64.

tsteven4:
The qtwebengine extension for WoA appeared in 6.10.0. The qtpdf extension for WoA appeared in 6.9.2.
Support for both the cross compiled and native versions is included and tested.

### DIFF
--- a/aqt/archives.py
+++ b/aqt/archives.py
@@ -425,8 +425,10 @@ class QtArchives:
 
         if self.version >= Version("6.8.0"):
             arch = self.arch
-            if self.os_name == "windows" or self.os_name == "windows_arm64":
+            if self.os_name == "windows":
                 arch = self.arch.replace("win64_", "", 1).replace("_cross_compiled", "", 1)
+            if self.os_name == "windows_arm64":
+                arch = self.arch.replace("win64_", "", 1)
             elif self.os_name == "linux":
                 arch = "x86_64"
             elif self.os_name == "linux_arm64":
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -294,8 +294,10 @@ class ArchiveId:
 
     def to_extension_folder(self, module, version, arch) -> str:
         extarch = arch
-        if self.host == "windows" or self.host == "windows_arm64":
+        if self.host == "windows":
             extarch = arch.replace("win64_", "", 1).replace("_cross_compiled", "", 1)
+        elif self.host == "windows_arm64":
+            extarch = arch.replace("win64_", "", 1)
         elif self.host == "linux":
             extarch = "x86_64"
         elif self.host == "linux_arm64":


Kidev:
I think we should split it in two parts instead of `or`ing the two cases, because it creates this output when trying to install a cross compiled version on host `windows_arm64`: 
```console
kidev:~$ aqt install-qt windows_arm64 desktop 6.10.0 win64_msvc2022_arm64_cross_compiled -m qtwebengine qtpdf --autodesktop

INFO    : aqtinstall(aqt) v3.2.2.dev39 on Python 3.13.7 [CPython GCC 15.2.1 20250813]
INFO    : Found extension qtwebengine
INFO    : Found extension qtpdf
ERROR   : The packages ['qt_base', 'qtpdf', 'qtwebengine'] were not found while parsing XML of package information!
``` 
By splitting like so:
```python
// metadata.py
if self.host == "windows" or self.host == "windows_arm64":
    extarch = arch.replace("win64_", "", 1)
    if  self.host == "windows":
        extarch = arch.replace("_cross_compiled", "", 1)
```
```python
// archives.py
if self.os_name == "windows" or self.os_name == "windows_arm64":
    arch = self.arch.replace("win64_", "", 1)
    if self.os_name == "windows":
        arch = self.arch.replace("_cross_compiled", "", 1)
```

We get:
```console
kidev:~$ aqt install-qt windows_arm64 desktop 6.10.0 win64_msvc2022_arm64_cross_compiled -m qtwebengine qtpdf --autodesktop

INFO    : aqtinstall(aqt) v3.2.2.dev39 on Python 3.13.7 [CPython GCC 15.2.1 20250813]
ERROR   : The packages ['qt_base', 'qtpdf', 'qtwebengine'] were not found while parsing XML of package information!
```

Which is less confusing imo

aqt/archives.py
```diff
@@ -425,8 +425,8 @@ def _get_archives_base(self, name, target_packages):
 
         if self.version >= Version("6.8.0"):
             arch = self.arch
-            if self.os_name == "windows":
-                arch = self.arch.replace("win64_", "", 1)
+            if self.os_name == "windows" or self.os_name == "windows_arm64":
+                arch = self.arch.replace("win64_", "", 1).replace("_cross_compiled", "", 1)
```
Kidev, line 429:
```suggestion
                arch = self.arch.replace("win64_", "", 1)
                if self.os_name == "windows":
                    arch = self.arch.replace("_cross_compiled", "", 1)
```

tsteven4:
@kidev I note that win64_msvc2022_arm64_cross_compiled does not show as an architecture for windows_arm64.

(aqttest-venv) tsteven4@PEDALDAMNIT:~/work/aqtinstall/dist$ !2359
aqt list-qt windows desktop --arch 6.10.0
win64_llvm_mingw win64_mingw win64_msvc2022_64 win64_msvc2022_arm64_cross_compiled
(aqttest-venv) tsteven4@PEDALDAMNIT:~/work/aqtinstall/dist$ aqt list-qt windows_arm64 desktop --arch 6.10.0
win64_msvc2022_arm64

Kidev:
Yes it does not exist, _cross_compiled version are only for `windows`, never `windows_arm64` (since their job is to replace the `windows_arm64` builds).
But by not splitting in two parts, `aqtinstall` always removes the `_cross_compiled` part, even on `windows_arm64`, and therefore confusingly shows
```
INFO    : Found extension qtwebengine
INFO    : Found extension qtpdf
``` 
Before failing correctly, mentionning that it could not find `qtwebengine` and `qtpdf`

aqt/metadata.py
```diff
@@ -294,8 +294,8 @@ def to_os_arch(self) -> str:
 
     def to_extension_folder(self, module, version, arch) -> str:
         extarch = arch
-        if self.host == "windows":
-            extarch = arch.replace("win64_", "", 1)
+        if self.host == "windows" or self.host == "windows_arm64":
+            extarch = arch.replace("win64_", "", 1).replace("_cross_compiled", "", 1)
```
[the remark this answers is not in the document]
tsteven4: A slightly different fix is in 4a53344c99a2db45c06e62e3c39510dcabc53ca7

aqt/archives.py
tsteven4: A slightly different fix is in 4a53344c99a2db45c06e62e3c39510dcabc53ca7

tsteven4:
I have successfully tested this in https://github.com/tsteven4/gpsbabel/tree/qt610test1 with github actions using a locally produced wheel.  We need these changes for our build on WoA with Qt 6.10.0.  We would like to pick up the fix from this PR in a new aqtinstall release for an upcoming release of gpsbabel.

PR #943, opened 2025-09-23
Bugfix for installing commercial Qt on Apple Silicon

oestej:
When using the official Qt installer on an Apple Silicon platform, the architecture is autodetected as mac-arm64, which fails, as the commercial Qt installer only provides a universal binary under mac-x64.

This PR falls back to mac-x86 for mac-arm64 if no build for mac-arm64 exists (like with the official commercial installer), as Qt seems to use mac-x86 for universal binary naming.


Kidev:
Duplicate #913 I think

PR #993, opened 2026-01-16
Add support to use config from environment variables

### LINKED ISSUE #992: [Feature]: Support AQT_MIRROR as environment variable
### Is your feature request related to a problem?

Thanks for the installer!
We uses this internally for the CI. Since our CI is behind the company firewall we want to use a specific mirror for aqt.

### The solution

The easiest way to provide that to the container builds is via environment variable. So it would be very helpful if AQT accept something like AQT_MIRROR as variable to use that mirror if no mirror is given via cmdline.

### Alternatives you've considered

_No response_

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

misery:
This is particularly helpful for short-lived containers in CI pipelines, where provisioning and mounting separate config files can be inconvenient. Using environment variables makes setups simpler, more portable, and easier to automate in container-based workflows.

Fixes #992

### DIFF
--- a/aqt/helper.py
+++ b/aqt/helper.py
@@ -443,8 +443,19 @@ def xml_to_modules(
 
 
 class MyConfigParser(ConfigParser):
+    def __init__(self) -> None:
+        super().__init__()
+        self.acceptable_env_keys = [
+            "MIRRORS_TRUSTED_MIRRORS",
+            "MIRRORS_BLACKLIST",
+            "MIRRORS_FALLBACKS",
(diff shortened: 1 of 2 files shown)


misery:
Yeah, fair enough. I followed your idea and added env variables to mirror handling only.

miurahr:
Thank you for updating manual description. Perfect!

misery:
Ready to merge?

PR #1000, opened 2026-03-01
Support Qt 6.11+ for Windows x64

### LINKED ISSUE #959: Add Qt 6.11.x support (Qt Repo folder structure changed)
There is no second `qt6_6110` folder contained within the first `qt6_6110` folder, Qt Repo has changed folder structure from `Qt 6.11`.

* See old folder structure containing a second `qt6_xxxx` folder.
https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_6101/

* New folder structure from `Qt 6.11`.
https://download.qt.io/online/qtsdkrepository/windows_x86/desktop/qt6_6110/
____

```
INFO    : aqtinstall(aqt) v3.3.0 on Python 3.12.10 [CPython MSC v.1943 64 bit (AMD64)]
WARNING : Failed to download checksum for the file 'online/qtsdkrepository/windows_x86/desktop/qt6_6110/qt6_6110/Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '6.11.0'.
* Please use 'aqt list-qt windows desktop' to show versions available.
Error: Error: The process 'C:\hostedtoolcache\windows\Python\3.12.10\x64\python.exe' failed with exit code 1
```

jdpurcell:
Fixes #959.

Notes:
1. I had to make `extension_for_arch` take an exact version number instead of `is_version_ge_6: bool`. Most callers already know the version, but a few don't, so I had to pass in fake versions. I don't like this but I don't understand the code well enough to come up with anything better.
2. I did a massive cleanup of `to_folder` even though it didn't really need any changes to support Qt 6.11. We can drop this commit if it's too risky. But I'm quite confident the existing code there is far more complicated than it needs to be.

### DIFF
--- a/aqt/metadata.py
+++ b/aqt/metadata.py
@@ -527,18 +527,25 @@ class QtRepoProperty:
                 return ext
         return ""
 
+    # Legacy folder structure - all arches in one folder
+    ANDROID_SHARED_FOLDER_VER = Version("5.15.2")
+    # Modern folder structure - separate folder for each arch
+    ANDROID_SPLIT_FOLDER_VER = Version("6.0.0")
+
     @staticmethod
     def possible_extensions_for_arch(arch: str) -> List[str]:
         """Assumes no knowledge of the Qt version"""
 
-        # ext_ge_6: the extension if the version is greater than or equal to 6.0.0
-        # ext_lt_6: the extension if the version is less than 6.0.0
-        ext_lt_6, ext_ge_6 = [
-            QtRepoProperty.extension_for_arch(arch, version) for version in (Version("5.15.0"), Version("6.0.0"))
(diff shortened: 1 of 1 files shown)


jdpurcell:
I figured out what the old code was doing when it passed in constant `true`/`false` values for `is_version_ge_6`. It was for version resolution where it does a folder listing and tries to locate relevant folders. Android is the only case that needs it so I put descriptive names instead of passing "magic" version numbers. Such code isn't needed for Windows x64 Qt 6.11 since Qt moved to the split folder structure starting with 6.8 - the folder listing doesn't contain extensions at this level anymore since they are now nested into subfolders.

